Issue #4, opened 2015-06-14
libextl won't compile with lua5.3 - change in lua_sethook signature.


jsbackus:
The signature for lua_sethook has changed in Lua 5.3 and no longer returns a value. Since lua_sethook is now a void, libextl no longer compiles.

raboof:
Interestingly, this function always returned '1', at least since 5.0.3 (http://www.lua.org/source/5.0/ldebug.c.html)

jsbackus:
From what I was able to gather, I don't think it has returned anything useful for a while. Looks like I may have chosen the wrong value... :smile: My fix worked, so I wonder if it is ever checked?

raboof:
Indeed it doesn't appear so - just removed the return value entirely :)

jsbackus:
Even better! Clearly if it compiled w/o a problem then it wasn't used. :smile: The good news is that it otherwise compiled fine under gcc 5, which has been quite a problem for some packages. Then again, it doesn't sound like C got much love in 5...

PR #5, opened 2015-06-14
Updated extl_sethook and extl_resethook to return 0.

jsbackus:
Fixes issue #4.


raboof:
(alternative fix already applied to master - still thanks for noticing!)

PR #6, opened 2015-06-28
mod_sm: Return NULL from match_cwin if XGetClassHint fails

JoeNotCharles:
Pillars of Eternity from gog.com causes notion to crash when it starts. The cause is that XGetClassHint returns 0, which means the "clss" output parameter is left pointing to random memory. Notion doesn't check the return value so it tries to access fields in clss and crashes.

Not sure why XGetClassHint fails. Possibly because the game goes fullscreen automatically on start.

### DIFF
--- a/mod_sm/sm_matchwin.c
+++ b/mod_sm/sm_matchwin.c
@@ -176,9 +176,9 @@ bool mod_sm_have_match_list()
 
 static WWinMatch *match_cwin(WClientWin *cwin)
 {
-    WWinMatch *match=match_list;
+    WWinMatch *match=NULL;
     int win_match;
-    XClassHint clss;
+    XClassHint clss = { NULL, NULL };
     char *client_id=mod_sm_get_client_id(cwin->win);
     char *window_role=mod_sm_get_window_role(cwin->win);
     char *wm_cmd=mod_sm_get_window_cmd(cwin->win);
@@ -191,19 +191,17 @@ static WWinMatch *match_cwin(WClientWin *cwin)
     
     if (!XGetClassHint(ioncore_g.dpy, cwin->win, &clss)) {
         warn("XGetClassHint failed");
-        return NULL;
+        goto done;
     }
 
-    for(; match!=NULL; match=match->next){
+    for(match=match_list; match!=NULL; match=match->next){
         
         win_match=0;
         
-        if(client_id || match->client_id){
-            if(xstreq(match->client_id, client_id)){
-                win_match+=2;
-                if(xstreq(match->window_role, window_role))
-                    win_match++;
-            }
+        if(xstreq(match->client_id, client_id)){
+            win_match+=2;
+            if(xstreq(match->window_role, window_role))
+                win_match++;
         }
         if(win_match<3){
             if(xstreq(match->wclass, clss.res_class) && xstreq(match->winstance, clss.res_name)){
@@ -211,8 +209,7 @@ static WWinMatch *match_cwin(WClientWin *cwin)
                 if(win_match<3){
                     if(xstreq(match->wm_cmd, wm_cmd))
                         win_match++;
-                    if(wm_name!=NULL && *wm_name!=NULL && 
-                       xstreq(match->wm_name, *wm_name)){
+                    if(wm_name!=NULL && xstreq(match->wm_name, *wm_name)){
                         win_match++;
                     }
                 }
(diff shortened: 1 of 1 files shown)


mod_sm/sm_matchwin.c
```diff
@@ -189,8 +189,11 @@ static WWinMatch *match_cwin(WClientWin *cwin)
     if(n<=0)
         assert(wm_name==NULL);
     
-    XGetClassHint(ioncore_g.dpy, cwin->win, &clss);
-    
+    if (!XGetClassHint(ioncore_g.dpy, cwin->win, &clss)) {
+        warn("XGetClassHint failed");
+        return NULL;
```
raboof: Doesn't this leak memory?

raboof:
Thanks for the report, looks like we should indeed check the `XGetClassHint` return value.

The current implementation also doesn't free res_name and res_class from the XClassHint, that seems like a leak, too. Haven't checked, though. 

Also, could you put in writing somewhere that you're OK with releasing your contributions under the LGPL as well as the Ion License? Thanks!

mod_sm/sm_matchwin.c
JoeNotCharles: Good point.

More issues I found:
- Yes, clss.res_name and clss.res_class leak.
- When Pillars of Eternity starts, client_id, window_role and wm_cmd all get set to NULL. (All those mod_sm_get_foo functions return NULL.) Accesses to client_id below are guarded by a null check, but the others aren't.
- According to http://tronche.com/gui/x/xlib/display/XFree.html, "A NULL pointer cannot be passed to this function." XFree(client_id), etc are called unconditionally which might be unsafe.

JoeNotCharles:
Grr, I commented before I pushed that second patch but it seems to have vanished.

You're right, clss leaked, and my first patch added a new leak. I also noticed another issue: the three mod_sm_get_* calls at the top of the func can return null. (I assume under the same circumstances that XGetClassHint fails.) When that happens the XFree and XFreeStringList calls are dangerous because according to http://tronche.com/gui/x/xlib/display/XFree.html: "A NULL pointer cannot be passed to this function." (Unlike the regular free func.) The second patch fixes both those issues.

I thought at first that the uses of client_id, window_role, wm_cmd were unsafe too because client_id was null-checked but the others weren't. But then I realized xstreq includes a null check already. So I removed the redundant null checks because I found them more confusing than helpful.

I give permission to release these contributions under the terms in the LICENSE file.

raboof:
Your previous comment is hidden under "raboof commented on an outdated diff" - yeah github is a bit wierd about that ;)

mod_sm/sm_matchwin.c
```diff
@@ -176,9 +176,9 @@ bool mod_sm_have_match_list()
 
 static WWinMatch *match_cwin(WClientWin *cwin)
 {
-    WWinMatch *match=match_list;
+    WWinMatch *match=NULL;
     int win_match;
-    XClassHint clss;
+    XClassHint clss = { NULL, NULL };
```
raboof: Nitpicking: some whitespace could be removed, e.g. `clss={NULL, NULL}` (http://notion.sourceforge.net/notionnotes/node7.html#SECTION00071000000000000000)

mod_sm/sm_matchwin.c
```diff
@@ -189,18 +189,19 @@ static WWinMatch *match_cwin(WClientWin *cwin)
...
         
         win_match=0;
         
-        if(client_id || match->client_id){
-            if(xstreq(match->client_id, client_id)){
-                win_match+=2;
-                if(xstreq(match->window_role, window_role))
-                    win_match++;
-            }
+        if(xstreq(match->client_id, client_id)){
```
raboof: AFAICS this now matches when both client_id and match->client_id are NULL, where it previously did not. Not sure what kind of effects that might have?

raboof:
Where does it guarantee that xstreq always does a null check?

raboof:
Regarding licensing, sorry to be a bit nitpicky there, but: if we'd ever want to release a version of Notion under the LGPL, can we have your permission to also include this patch?

mod_sm/sm_matchwin.c
JoeNotCharles: #define xstreq(a,b) (a && b && (strcmp(a,b)==0))

So if either or both are null it returns false, just like before.

raboof: Ah didn't realize this is a local definition, looks good great!

JoeNotCharles:
Yes, you may release this patch under the LGPL.

raboof:
Thanks so much! Merged!

PR #8, opened 2015-07-04
* procps has changed the output of free (and removed -o).

jsbackus:
Patch detects version of free and acts accordingly.

### DIFF
--- a/contrib/statusd/statusd_mem.lua
+++ b/contrib/statusd/statusd_mem.lua
@@ -67,27 +67,27 @@ local free_version_minor
 
 local function show_meminfo(status)
 	while status do	
-                local buffers, cached, avail, hused
-	        local ok, _, total, used, free, shared, col5, col6 =--
+		local buffers, cached, avail, hused
+		local ok, _, total, used, free, shared, col5, col6 =--
 		string.find(status, "Mem:%s+(%d+)%s+(%d+)%s+(%d+)%s+(%d+)%s+(%d+)%s+(%d+)")
 		--
 		if not ok then statusd.inform("mem_template", "--") return end
 		--
 		if (3.3 <= free_version_major) and (10 <= free_version_minor) then
-		  -- In 3.3.10 and later, there is no separate cached column. Instead it is
-		  -- the total amount available. So...
+			-- In 3.3.10 and later, there is no separate cached column. Instead it is
+			-- the total amount available. So...
 
-		  -- Total available is simply column 6
-		  avail = col6
-		  -- Use the buffers amount for cached
-		  buffers = col5
-		  cached = col5
-		  -- Total used is now total - available
-		  hused = total - avail
+			-- Total available is simply column 6
+			avail = col6
+			-- Use the buffers amount for cached
+			buffers = col5
+			cached = col5
+			-- Total used is now total - available
+			hused = total - avail
 		else
-		  buffers = col5
-		  cached = col6
-		  avail = free + cached + buffers
+			buffers = col5
+			cached = col6
+			avail = free + cached + buffers
 		end
 		--
 		statusd.inform("mem_total", total)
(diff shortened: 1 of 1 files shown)


contrib/statusd/statusd_mem.lua
```diff
@@ -75,9 +77,23 @@ local function show_meminfo(status)
 		statusd.inform("mem_free", free)
 		statusd.inform("mem_shared", shared)
 		statusd.inform("mem_buffers", buffers)
-		statusd.inform("mem_cached", cached)
-		statusd.inform("mem_hused", tostring(used - cached - buffers))
-		statusd.inform("mem_hfree", tostring(free + cached + buffers))
+		--
+		if (3.3 <= free_version_major) and (10 <= free_version_minor) then
+		  local avail = cached
```
raboof: Hmm, perhaps we should rename `cached` above to something more generic?

raboof: and `buffers`, too, I guess.

contrib/statusd/statusd_mem.lua
```diff
@@ -93,10 +109,32 @@ local function show_meminfo(status)
...
+      
+      -- Run free with the -V option to get its version information
+      fh = io.popen("free -V", "r")
+      version_info = fh:read()
+      fh.close()
+      
+      ok, _, free_version_major, free_version_minor =--
+        string.find(version_info, "(%d+%.%d+)%.(%d+)")
+
+      if not ok then statusd.inform("mem_template", "--") return end
```
raboof: Perhaps when we can't determine the version of 'free' we should fall back to the old behavior?

contrib/statusd/statusd_mem.lua
jsbackus: Good point. I've cleaned it up and pushed an update. Now to figure out how to update a pull request..... :smile:

contrib/statusd/statusd_mem.lua
jsbackus: Gah, yeah, that was the intention. Got it right in show_meminfo. Completely missed it in get_free_version. :smile:

jsbackus:
Addressed each concern in separate commits to hopefully make it easier to pick and choose. I'm happy to continue making tweaks, just let me know.

contrib/statusd/statusd_mem.lua
```diff
@@ -60,43 +60,90 @@
...
 local settings = table.join(statusd.get_config("mem"), defaults)
+local free_version_major
+local free_version_minor
 
 ------- MEM MONITOR :----------------------------------------------------------------------
 
 local function show_meminfo(status)
 	while status do	
-		local ok, _, total, used, free, shared, buffers, cached =--
+                local buffers, cached, avail, hused
```
raboof: Please use tabs instead of spaces (as in the rest of this file) (or change the whole file to spaces ;) )

contrib/statusd/statusd_mem.lua
```diff
@@ -60,43 +60,90 @@
...
+	        local ok, _, total, used, free, shared, col5, col6 =--
 		string.find(status, "Mem:%s+(%d+)%s+(%d+)%s+(%d+)%s+(%d+)%s+(%d+)%s+(%d+)")
 		--
 		if not ok then statusd.inform("mem_template", "--") return end
 		--
+		if (3.3 <= free_version_major) and (10 <= free_version_minor) then
+		  -- In 3.3.10 and later, there is no separate cached column. Instead it is
```
raboof: Here, too

contrib/statusd/statusd_mem.lua
jsbackus: :innocent: Yeah, sorry - Emacs "helped" me. Kept the tabs, since it was already using tabs. Please let me know if I missed something else. This coding past my bedtime is dangerous... :smiley:

contrib/statusd/statusd_mem.lua
```diff
@@ -60,43 +60,90 @@
...
+			-- In 3.3.10 and later, there is no separate cached column. Instead it is
+			-- the total amount available. So...
+
+			-- Total available is simply column 6
+			avail = col6
+			-- Use the buffers amount for cached
+			buffers = col5
+			cached = col5
+			-- Total used is now total - available
+			hused = total - avail
```
raboof: Shouldn't the assignment `hused = total - avail' be outside the if/then/else?

jsbackus: doh! Good catch. No, I just forgot to add the corresponding else statement, which is: `hused = used - cached - buffers`

The *should* be synonymous, but just to be sure... I'll fix it this evening.

jsbackus: Managed to find a machine I could actually run the code on with an older version of free. Found another bug, which I've fixed. I'll push the changes tonight.

contrib/statusd/statusd_mem.lua
```diff
@@ -60,43 +60,91 @@
...
+		else
+			buffers = col5
+			cached = col6
+			avail = free + cached + buffers
+			hused = used - cached - buffers
```
jsbackus: hused now properly set in both clauses of if...

contrib/statusd/statusd_mem.lua
```diff
@@ -60,43 +60,91 @@
...
 		statusd.inform("mem_used", used)
 		statusd.inform("mem_free", free)
 		statusd.inform("mem_shared", shared)
 		statusd.inform("mem_buffers", buffers)
 		statusd.inform("mem_cached", cached)
-		statusd.inform("mem_hused", tostring(used - cached - buffers))
-		statusd.inform("mem_hfree", tostring(free + cached + buffers))
+		statusd.inform("mem_hused", string.format('%0d', hused))
+		statusd.inform("mem_hfree", string.format('%0d', avail))
 		--
```
jsbackus: Changed these two lines to use appropriate value from if/else. Also changed to string.format() because Lua 5.3 has added a distinction between integers and floats, so now tacks on a '.0' when converting numbers to strings. Should work fine in 5.2, but it'll be tomorrow before I can verify.

jsbackus:
Thanks for your patience!

contrib/statusd/statusd_mem.lua
raboof: Changes look good, will merge when your test with 5.2 is succesful!

Thanks for iterating - I'm really liking how github makes that easy.

jsbackus: Changes worked.

Welcome! Yeah, this flow is very nice. Particularly being able to comment in-line.

PR #12, opened 2015-08-19
New statusd_inetaddr2 statusbar template

knixeur:
It displays selected interfaces's IPv4 and IPv6 address in the
statusbar.

### DIFF
--- a/contrib/statusd/statusd_inetaddr2.lua
+++ b/contrib/statusd/statusd_inetaddr2.lua
@@ -1,0 +1,86 @@
+-- Authors: Guillermo Bonvehí <gbonvehi@gmail.com>
+-- License: Public domain
+--
+-- statusd_inetaddr2.lua -- show IP address for Notion's statusbar
+
+-- Shows assigned IP address for each configured interface.
+
+-- Uses /sbin/ifconfig to fetch current iface's data
+
+-- Configuration:
+
+-- `template' string controls how the output will be formated.  It may
+-- contain the following keywords:
+
+-- PLACEHOLDER  DESCRIPTION             EXAMPLE
+-- -----------  -----------             -------
+-- iface        interface               eth0
+-- ipv4         IPv4 address            192.168.1.1
+-- ipv6         IPv6 address              ::1
+
+-- `interfaces' table contains the interfaces to be monitored.
+-- `separator' is placed between the entries in the output.
+-- `interval' is the refresh timer (in milliseconds).
+
+-- This software is in the public domain.
+
+--------------------------------------------------------------------------------
+
+local defaults = {
+    template = "[%iface %ipv4 %ipv6]",
+    interfaces = { "eth0", "wlan0" },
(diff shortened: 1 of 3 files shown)


gbonvehi:
I need to review how to parse the IP address as different ifconfig versions have different output.

jeffpc:
It gets worse... ifconfig output can be *very* different on other UNIX systems.  E.g., from OpenIndiana (an illumos distro):

```
$ ifconfig 
lo0: flags=2001000849<UP,LOOPBACK,RUNNING,MULTICAST,IPv4,VIRTUAL> mtu 8232 index 1
	inet 127.0.0.1 netmask ff000000 
dhcp0: flags=1004843<UP,BROADCAST,RUNNING,MULTICAST,DHCP,IPv4> mtu 1500 index 2
	inet 10.10.1.34 netmask fffffe00 broadcast 10.10.1.255
```

OI has no `ip` command either.

raboof:
Thanks for your contribution!

I don't have a big problem with accepting a 'contrib' script that does not work on all platforms, but perhaps we should specify which platforms are known to work at the top of the script?

Given that tool output may change even within a platform, it might be good to add a testcase to make sure future changes to this script won't break the output as it appears on your system. On the other hand we don't really have tests for the contrib scripts yet, so this might be a bit of extra work - it's up to you whether you'd like to take this on. mod_xinerama and mod_xrandr do have some tests you could use as inspiration.

wilhelmy:
Just a note, I think some distros have ip(8) in /usr/sbin.

raboof:
I see the latest commit here was 'WIP', can you ping when this is ready to go?

knixeur:
Sure, I forgot about this PR but still use it. I'll try to make it more generic by detecting which tool is available.

contrib/statusd/statusd_inetaddr2.lua
```diff
@@ -38,21 +38,23 @@ local inetaddr2_timer = statusd.create_timer()
...
     end
     while true do
         local line = f:read("*l")
         if line == nil then break end
         if (string.match(line, inet4_match)) then
-            inet4_addr = line:match("inet addr:(%d+%.%d+%.%d+%.%d+)")
-        elseif (string.match(line, inet4_match)) then
-            inet6_addr = line:match(("([a-fA-F0-9]*):"):rep(8):gsub(":$","$"))
+            inet4_addr = line:match("inet (%d+%.%d+%.%d+%.%d+)")
+            if inet4_addr == nil then inet4_addr = "none" end
```
wilhelmy: I think this can shrink to `inet4_addr = line:match(...) or "none"`

knixeur:
I think this is ready now, you can choose to use ip addr or ifconfig and it will search for the binary in /sbin/ /usr/sbin/ /bin/, etc.
I could also use PATH environment value, but I think it's usable at it is.

raboof:
Nice!

Issue #9, opened 2015-08-08
Use pkg-config for all library stuff


xmw:
Hi,
I took a look at the updated Gentoo package and wrote a patch [1] to use pkg-config for all library flags.
It would be nice to have on installations with awkward library location in general, i.e. Gentoo/Prefix.
Thanks
[1] https://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo-x86/x11-wm/notion/files/notion-3_p2015061300-pkg-config.patch?view=markup

raboof:
Should now be fixed via #13 , thanks for your report and patch (and thanks to @knixeur for the PR :) )

PR #13, opened 2015-08-19
Closes issue #9 - Apply patch suggested by @xmw

knixeur:
The patch changes declaratiosn to use pkg-config instead of hardcoded
flags as suggested by @xmw
See: 
- https://github.com/raboof/notion/issues/9
- https://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo-x86/x11-wm/notion/files/notion-3_p2015061300-pkg-config.patch?view=markup


raboof:
Thanks for your contribution, sorry for taking so long!

PR #17, opened 2016-04-26
Add sizepolicy support for menus

hedning:
This allows you to position menus according to sizepolicy. Which lets you use center for instance, which is often nicer than the lower-left corner. I've been using the patch for a long time without any issues.

### DIFF
--- a/mod_menu/grabmenu.c
+++ b/mod_menu/grabmenu.c
@@ -93,7 +93,8 @@ WMenu *mod_menu_do_grabmenu(WMPlex *mplex, ExtlFn handler, ExtlTab tab,
                MPLEX_ATTACH_LEVEL|
                MPLEX_ATTACH_UNNUMBERED|
                MPLEX_ATTACH_SIZEPOLICY);
-    par.szplcy=SIZEPOLICY_FULL_BOUNDS;
+    if (FALSE==extl_table_gets_sizepolicy(param, "sizepolicy", &par.szplcy))
+        par.szplcy=SIZEPOLICY_FULL_BOUNDS;
     par.level=STACKING_LEVEL_MODAL1+2;
 
     menu=(WMenu*)mplex_do_attach_new(mplex, &par,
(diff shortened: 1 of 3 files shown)


mod_menu/grabmenu.c
```diff
@@ -93,7 +93,7 @@ WMenu *mod_menu_do_grabmenu(WMPlex *mplex, ExtlFn handler, ExtlTab tab,
                MPLEX_ATTACH_LEVEL|
                MPLEX_ATTACH_UNNUMBERED|
                MPLEX_ATTACH_SIZEPOLICY);
-    par.szplcy=SIZEPOLICY_FULL_BOUNDS;
+    extl_table_gets_sizepolicy(param, "sizepolicy", &par.szplcy);
```
raboof: Doesn't this mean that when 'sizepolicy' is not set, the sizepolicy will be undefined?

hedning:
extl_table_gets_sizepolicy sets it to  SIZEPOLICY_DEFAULT via string2sizepolicy. This seems to be the lower-left corner for me, but it's somewhat unclear what the default actually is. It could be set explicitly to lower-left if the table_gets function returns false of course.

mod_menu/grabmenu.c
hedning: extl_table_gets_sizepolicy sets it to SIZEPOLICY_DEFAULT via string2sizepolicy. This seems to be the lower-left corner for me, but it's somewhat unclear what the default actually is. It could be set explicitly to lower-left if the table_gets function returns false of course.

raboof: AFAICS `extl_table_gets_sizepolicy` will not call `string2sizepolicy` when 'sizepolicy' is not set - indeed it will fall back to `SIZEPOLICY_DEFAULT` when an unknown value was encountered.

I think it'd be good to explicitly check whether `extl_table_gets_sizepolicy` returns -1, and in that case set the sizepolicy to `SIZEPOLICY_FULL_BOUNDS` (for maximum backwards compatibility).

hedning:
You're right, not sure why it has functioned without a proper default. It now defaults to full_bounds, as before.

raboof:
Yeah, C can be scary like that :).

Squashing the commits would be nice!

hedning:
Hehe, there we go.

raboof:
Thanks!

raboof:
@hedning for the record, could you please state you're OK with releasing your Notion contributions both under the Ion license and under the LGPL?

hedning:
Ah, yeah, I'm okay with releasing my contributions under both the Ion license and the LGPL. Do I have to put it somewhere else to?

raboof:
No this works for me, thanks!

PR #23, opened 2016-10-27
Center cursor warp to window center

ideasman42:
For large windows (using one window per monitor for eg), using the top corner is quite annoying.

Its common for other window managers to use window center (xmonad, spectrwm for eg).

### DIFF
--- a/ioncore/focus.c
+++ b/ioncore/focus.c
@@ -501,8 +501,8 @@ bool region_do_warp_default(WRegion *reg)
         if(px>=x && py>=y && px<x+w && py<y+h)
             return TRUE;
     }
-    
-    rootwin_warp_pointer(root, x+5, y+5);
+
+    rootwin_warp_pointer(root, x+w/2, y+h/2);
         
     return TRUE;
 }


raboof:
Hi @ideasman42 - I agree this seems nice, on the other hand I don't really want to interfere with the behavior existing users have come to expect: for example, some user that rarely use the mouse might be annoyed by having this mouse pointer so prominently on their windows.

It would be nice if this could be an option though - perhaps even as a lua script to include in your configuration?

raboof:
I found this in my personal configuration (though disabled):

```
local function warpToMiddle(reg)
  local root = reg:rootwin_of();
  local geom = geom_relative_to(reg, root);
  local x = geom.x + math.floor(geom.w / 2);
  local y = geom.y + math.floor(geom.h / 2);
  root:warp_pointer(x, y);
  return true;
end
notioncore.get_hook('region_do_warp_alt'):add(warpToMiddle);
```

ideasman42:
Hi, while it could be a hook, as you've suggested, it seems a bit annoying to use a hook for this?

I tried to use the hook but am missing `geom_relative_to`, seems this would be nice to avoid flickering cursor when focusing on windows.

Another reason to use the center is bumping the mouse accidentally can easily focus on another window.

----

Possibly this could be configurable? (pixel-padding, x-factor, y-factor).

So default could be: `pixel-padding=4, xfac=0.0, yfac=0.0` (top left).

Using the middle could use: `pixel-padding=0, xfac=0.5, yfac=0.5`.

Or have named positions (top-left, top-right, mid-left, mid-right, mid-mid... etc).

raboof:
added `geom_relative_to` to the example above.

I'm not sure I understand why a hook would be annoying and what 'flickering' you see?

ideasman42:
`geom_relative_to` is still missing from the example.

As for flickering, I couldn't test the hook, just assumed 2x mouse-warps would be noticeable.

raboof:
Oops forgot to hit 'update' on the example, sorry :/.

The default behavior is also achieved via the `region_do_warp_alt` hook (https://github.com/raboof/notion/blob/master/ioncore/ioncore.c#L305), so adding `warpToMiddle` to this `_alt` hook and returning `true` makes sure this logic is executed instead of the default behaviour, not after it. See http://notion.sourceforge.net/notionconf/node6.html#SECTION00620000000000000000 for a bit more on alt hooks.

ideasman42:
@raboof, checked the hook, its a little over active - for example switching between windows in the same frame also centers the cursor.

Probably this should check if the cursor is outside the frame before warping.

ideasman42:
Edit, submitted PR not to warp when the cursor is in the frame.

raboof:
Wouldn't it be useful to give the hook the power to decide whether it wants to move the cursor when it is already inside the frame? You can extend the hook not to warp in that case, right?

ideasman42:
@raboof, as stated in previous PR, its not impossible that someone might want this... though in practice I'm not sure why you would. Nevertheless point taken.

Part of the issue here *(unless I'm mistaken)* is that its not trivial to lean the API to write the Lua to figure out when this should run. Eg, to convert this C code to Lua.

Part of the reason Im suggesting this as an option - is this is something a user might want to change on first using notion, and to learn enough lua to do this isn't some 10min task (from what I can tell).

An alternative could be to have a pre-written snippet (on a wiki or so) people can paste into their config, that also checks if the cursor is in the frame.

raboof:
I agree this could be a common wish for new users, and we shouldn't require them to learn enough lua to write this function themselves - but indeed we could provide the ready-to-go script, which should also give them a taste of just how customizable notion is.

A script in `contrib` and perhaps a reference to it from some documentation somewhere might be a good place to start?

ideasman42:
Submitted PR #27, related to this one. (makes it redundant)

raboof:
Ok :+1:

PR #26, opened 2016-11-13
Don't call region_do_warp_alt callback when the cursor is in the frame

ideasman42:
Running the cursor warping callback when the cursor is in the frame is annoying and runs when:

- Clicking on tabs
- Cycling between frames (Alt-Tab)

### DIFF
--- a/ioncore/focus.c
+++ b/ioncore/focus.c
@@ -452,8 +452,9 @@ bool ioncore_should_focus_parent_when_refusing_focus(WRegion* reg){
 }
 
 void region_finalise_focusing(WRegion* reg, Window win, bool warp, Time time, int set_input) { 
-    if(warp)
+    if(warp && !region_is_cursor_inside(reg)){
         region_do_warp(reg);
+    }
     
     if(REGION_IS_ACTIVE(reg) && ioncore_await_focus()==NULL)
         return;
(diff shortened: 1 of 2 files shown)


raboof:
I'm not sure this is annoying - the hook can decide whether or not to warp the cursor.

ideasman42:
Right, the callback can do this.
Although I'm not sure what you would want clicking on a tab or Alt-tab to warp the mouse... 
It just seems such an unlikely use-case that it *could* be ignored.

raboof:
As I didn't find any cases of hooks that take advantage of this power I'm OK with moving this complexity out of the hook and into the C code.

I would like to see it moved to before `region_do_warp` is being called though, to make both the hook and `region_do_warp` itself simpler.

ideasman42:
Updated the patch so it checks cursor state before warping:

```
    if(warp && !region_is_cursor_inside(reg)){
        region_do_warp(reg);
    }
```

region_is_cursor_inside & region_do_warp_default duplicate some logic. but not much... could be extracted into a funciton though not sure its worth it (only 3x function calls)

raboof:
We could consider doing 
```
reg=find_warp_to_reg(reg);
     		     
     if(reg==NULL)		     if(reg==NULL)
         return FALSE;		         return FALSE;
```
exactly once, and pass the new `reg` to `region_do_warp` - what do you think?

ideasman42:
@raboof, not sure what you meant exactly, but finding the warp region once makes sense.

This changes behavior of `region_do_warp` if its called externally, (its not currently).
Possibly this could be made into 2x functions.

Where `region_do_warp` would become:

```
        WRegion* reg_warp=find_warp_to_reg(reg);
         if(reg_warp!=NULL && region_is_cursor_inside(reg_warp)){
            region_do_warp(reg_warp);
        }
```

raboof:
Cool, starts to look good!

I'm afraid we'll have to move `find_warp_to_reg` up (see the build failure on travis) now though :)

raboof:
An explicit declaration works too :).

Thanks!!

PR #27, opened 2016-11-16
Add options for warp cursor placement

ideasman42:
Add options to offset the cursor when warping, while this is possible using callbacks, typically users will want to select a corner or the center of the frame.

There are no changes to the default behavior, now users can easily center the cursor by changing some options.

From the embedded config:

```
    -- Margin to use when warping the cursor (in pixels),
    -- to inset from the edge of the frame.
    --warp_margin=5,

    -- Placement to use when positioning the cursor in the frame
    -- (as a factor from 0-1).
    -- Where an x,y value:
    -- * (0.0, 0.0) for top-left.
    -- * (1.0, 1.0) for bottom-right.
    -- * (0.5, 0.5) for the center of the frame.
    --warp_factor_x=0.0,
    --warp_factor_y=0.0,
```

---- 
This also corrects mistake in previous PR, `region_is_cursor_inside` check was flipped.

### DIFF
--- a/etc/cfg_notion.lua
+++ b/etc/cfg_notion.lua
@@ -42,6 +42,19 @@ ioncore.set{
     -- changing focus. Enabled by default.
     --warp=true,
 
+    -- Margin to use when warping the cursor (in pixels),
+    -- to inset from the edge of the frame.
+    --warp_margin=5,
+
+    -- Placement to use when positioning the cursor in the frame
+    -- (as a factor from 0-1).
+    -- Where an x,y value:
+    -- * (0.0, 0.0) for top-left.
+    -- * (1.0, 1.0) for bottom-right.
+    -- * (0.5, 0.5) for the center of the frame.
+    --warp_factor_x=0.0,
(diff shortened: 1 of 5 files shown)


etc/cfg_notion.lua
```diff
@@ -42,6 +42,19 @@ ioncore.set{
...
+    --warp_margin=5,
+
+    -- Placement to use when positioning the cursor in the frame
+    -- (as a factor from 0-1).
+    -- Where an x,y value:
+    -- * (0.0, 0.0) for top-left.
+    -- * (1.0, 1.0) for bottom-right.
+    -- * (0.5, 0.5) for the center of the frame.
+    --warp_factor_x=0.0,
+    --warp_factor_y=0.0,
```
raboof: Nice and well-documented

raboof:
While my personal preference would probably be to keep notions' C core as simple as possible and solve this with a well-documented (ready-to-go) hook, I really appreciate your contribution, so merging :)

Thanks!!

PR #28, opened 2016-11-16
Cleanup: 4 space indentation

ideasman42:
Some files mixed in 2 space indentation.
Left files alone which look to be imported from elsewhere `libtu/snprintf_2.2` for eg.

Also use newline for function braces


raboof:
thx!

PR #29, opened 2016-11-16
Cleanup: unused param warnings

ideasman42:


raboof:
For each of the parameters now marked `UNUSED`, are we sure they are intentionally unused?

Did you verify the parameter cannot be removed entirely for those cases, and/or that the parameter was actually intended to be used?

mod_sm/sm_session.c
```diff
@@ -143,16 +143,16 @@ static void sm_set_other_properties()
     char *rmarg="-rf";
     int nvals=0, i;
     const char *sdir=NULL, *cid=NULL;
-    
-    SmPropValue discard_val[3];
-    SmProp discard_prop={ SmDiscardCommand, SmLISTofARRAY8, 3, NULL };
+
+    /*SmPropValue discard_val[3];*/
```
raboof: I'm not a big fan of commenting out code: we should try and figure out what the reason for having this code was, and either fix the code (as obviously it currently has no effect) or remove it entirely.

raboof: interestingly I seem to have previously commented out this code myself in dd019523c8996503df609f299b839ce5966becf8, but then reverted that again in d9e5df32

ideasman42: First of all my aim is to have notion compiling with zero warnings.

Currently there are quite a few and it makes development not-so-nice,
I much prefer to have zero warnings where possible. And note TODO's in the code which can be handled later.

Notion is quite close to having no warnings, so I think its nice to quiet them for as long as its not causing large/disruptive changes to the code-base.

`UNUSED` parameters in the code could be considered implicit TODO's - at some point it makes sense to check why they're there and if they could be removed.
Quite a few cases touched in this patch should probably be kept, for API consistency.
eg: `end_scroll(WMenu *UNUSED(menu))`

In this particular case of `discard_prop` I don't have a strong opinion.

Commenting lines as I've done in this patch causes diff-noise, so it could be ifdef'd.

eg:

    // #define USE_DISCARD
    #ifdef USE_DISCARD
        SmPropValue discard_val[3];
        SmProp discard_prop={ SmDiscardCommand, SmLISTofARRAY8, 3, NULL };
    #endif

OTOH, if you're sure it can be removed... just remove it?

raboof: I agree having no warnings would be great.

I agree replacing warnings with TODO's would be a good idea where possible - when doing that, I think it's nice to have a link from the TODO to a GitHub issue where we can further discuss/track that particular TODO.

Actually `UNUSED` parameters are not implicit TODO's: sometimes it is actually necessary to have the parameter present in the signature, even though it is not actually used in that implementation - for example when working with 'interfaces'/'inheritance'/'hooks' based on function pointers.

I suppose we could make some aliases for `UNUSED` that distinguish the various cases (`UNUSED_TODO_CONSIDER_REMOVAL`, `UNUSED_BUT_KEPT_FOR_CONSISTENCY`? :) )

As for whether I'm sure it can be removed... I'm not, otherwise it'd have gone long ago :). Would be nice to really understand what's going on before finally removing it.

ideasman42: Could the `UNUSED` be applied in cases where its fairly clear that the argument should be kept for consistency?

I recall quite a few times the argument could be removed but IMHO it was better to keep and tag as unused. IIRC a handful of cases would be candidates for removal.

raboof: Yeah using `UNUSED` where it's for consistency makes sense to me!

raboof:
Continued in #75

PR #30, opened 2016-11-16
Correct missing parens and confusing logical op

ideasman42:


mod_sp/main.c
```diff
@@ -192,9 +192,10 @@ bool mod_sp_set_shown_on(WMPlex *mplex, const char *how)
 
     if(!found){
         int sp=libtu_string_to_setparam(how);
-        if(sp==SETPARAM_SET || sp==SETPARAM_TOGGLE)
+        if(sp==SETPARAM_SET || sp==SETPARAM_TOGGLE){
```
raboof: Good catch! A quick grep didn't reveal any places that used the return value of this function, so this seems like a safe change/fix.

PR #35, opened 2016-12-13
Cleanup: whitespace

ideasman42:
White space cleanup - trailing space for C source

### DIFF
--- a/de/brush.c
+++ b/de/brush.c
@@ -1,7 +1,7 @@
 /*
  * ion/de/brush.c
  *
- * Copyright (c) Tuomo Valkonen 1999-2009. 
+ * Copyright (c) Tuomo Valkonen 1999-2009.
  *
  * See the included file LICENSE for details.
  */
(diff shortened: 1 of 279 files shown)


de/brush.c
```diff
@@ -45,31 +45,31 @@ bool debrush_init(DEBrush *brush, Window win,
     brush->indicator_w=0;
     brush->win=win;
     brush->clip_set=FALSE;
-    
+
```
raboof: Hmm, are we sure we want to remove the indentation from these otherwise-empty lines? Did you use a particular tool to remove the trailing whitespace?

ideasman42: Sure, its mostly annoying on lines after text.

I have the impression that its becoming more common to consider any trailing space to be some kind of *lint*. (git flags it up as red in diffs, from what I can tell its common for developers to configure their editors to highlight trailing space, or even strip on save).

One reason to remove this, is a developer might press delete at the end of a line, without realizing they add trailing space (I've seen this happen a few times).

Personally I don't have a strong opinion, just prefer not to have trailing space.

Command used:

    find . -name '*.c' | xargs perl -pi -e 's/\s*$//'

raboof: OK, agreed!

Issue #38, opened 2017-01-31
libextl: extl_stack_get doesn't work for boolean values when ExtlAny is requested


olejorgenb:
I'm working on adding icons to tabs and when adding styling support I noticed that `show_icon = true` in a defstyle wasn't retrievable using `grbrush_get_extra`. 

Digging I see that the code initializing the brush extra table  ends up calling `extl_stack_get(st, -1, 'a', FALSE, NULL, &v)` (thorugh `extl_table_iter_do`). Ie. get the stack value as a ExtlAny. But `extl_stack_get` is missing a case when the actual stack type is `LUA_TBOOLEAN`.

It's simple enough to fix:

        case LUA_TBOOLEAN:
            if(type!='b' && type!='a')
                return FALSE;
    
            b=lua_toboolean(st, pos);
    
            if(type=='b'){
                if(valret)
                    *((bool*)valret)=b;
            }else if(type=='a'){
                if(valret){
                    ((ExtlAny*)valret)->type='b';
                    ((ExtlAny*)valret)->value.b=b;
                }
            }
            return TRUE;

I assume this case simply was forgotten (or maybe ExtlAny didn't support boolean at some point). Correct if there's an actual reason for the omission.

**If** I complete the icon branch **and** it's accepted, the fix will be included there. Otherwise the bug's at least exposed here :)

raboof:
I'm not aware of any reason for the omission, thanks for recording the situation here!

knixeur:
This was fixed by merging #79 thanks @olejorgenb !

PR #39, opened 2017-02-05
WIP: Tab icons

olejorgenb:
Work in progress adding icons to tabs. (ie. render `_NET_WM_ICON`)

Status: Works, but the code need some cleaning, and I haven't decided on all design choices yet.

Toggled through a style attribute `show_icon`. It's enabled for the clean looks in the branch. If you use a different look (or keep the look files in ~/.notion) add  `show_icon = true` to `defstyle("tab", ...` or `defstyle("tab-frame", ...`

`icon_align_left=false|true` controls the alignment (whether the icon follows the text of is always left aligned)

I put this PR out there a bit early to get some indication on whether the feature is wanted at all and maybe some feedback/input. If there's little interest I might just maintain the branch in a less polished state.

cairo is a dependency.

1. Is there strong feelings about needing a build flag? (a runtime option to will obviously be included)
2. Is it worth adding `icon_path` to winprops?
3. The icon size is hadcoded to 16x16 px atm. Is anyone interested in larger/smaller sizes?

(ref: #38)

2017-02-22-182019_992x676_scrot

### DIFF
--- a/de/draw.c
+++ b/de/draw.c
@@ -551,8 +551,9 @@ void debrush_draw_iconbox(DEBrush *brush,
     debrush_do_draw_box(brush, geom, cg, needfill);
 
     if(elem->icon){
-        /* TODO: Probably not the best idea to lie about the real window dimensions */
-        cairo_surface_t *xsurf=cairo_xlib_surface_create(ioncore_g.dpy, brush->win, DefaultVisual(ioncore_g.dpy, 0), geom->w, geom->h);
+        /* TODO: Probably not the best idea to lie about the real window dimensions (but under reporting seems to work) */
+        /* TODO: How expensive is it to create a new surface per icon? Could easily reuse it */
+        cairo_surface_t *xsurf=cairo_xlib_surface_create(ioncore_g.dpy, brush->win, DefaultVisual(ioncore_g.dpy, 0), geom->x+geom->w, geom->h);
         cairo_t *cr = cairo_create(xsurf);
         cairo_set_source_surface(cr, elem->icon, geom->x+bdw.left+1, geom->y+bdw.top);
         /* cairo_rectangle(cr, geom->x, geom->y, 16, 16); */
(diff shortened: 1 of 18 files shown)


raboof:
Cool!

I think adding cairo as a build-time required dependency could be a hurdle, it'd be really nice if that could be avoided...

Notion users don't tend to like changes to their configurations, but leaving this disabled by default will make it hard to discover. Not sure how to handle that gracefully...

olejorgenb:
It's simple enough (not pretty) to wrap all icon stuff in `#ifdefs`. I'm sure it's possible to make do with only Xlib draw functions also, but I'm not particular motivated to go down that road. If someone is willing to work out the details it's cool though.

We could extend `~/.notion/.welcome_msg_displayed` to include version information and use that to display a "what's new" dialog but not sure it's worth the complexity.

wilhelmy:
Hypothetically speaking, how hard would it be to ditch cairo (since it isn't used elsewhere in notion right now)? Then nothing would prevent this being merged.

olejorgenb:
A long time since I wrote this, but I don't think it's any harder than to find something else that can render ARGB data to a x-drawable. If you want support for the icon_path winprop you'll need something that can decode png  - hopefully to ARGB laid out like the _NET_WM_ICON xprop (to simplify buffer managment)

Probably possible to use https://tronche.com/gui/x/xlib/utilities/XCreateImage.html though I'm not sure how well it handles alpha (I don't actually remember if I tried to use the xlib functions or not)

knixeur:
Thanks for the info. The code is very clear we could ifdef the code if Cairo is present like we do with Xft

olejorgenb:
PS: although the PR says WIP, me and hedning used the patch daily (for a year+) as part of our ["fork"](https://github.com/paperwm/notion), so the code should be reasonably well tested.

knixeur:
Great! Thanks @olejorgenb I'll rebase against master (instead of notion3) and create new PR

knixeur:
I'll close this to follow it on #211 , thanks again @olejorgenb !

PR #42, opened 2017-04-08
notionflux: distinct exit status on lua error

olejorgenb:
At least three useful groups of notionflux results
1. Technical failure (notion not running, to much data to send, etc.)
2. Error when loading or running lua code (syntax error, NPE, etc.)
3. Success

It's possible to differentiate between these today since the output for case 2 is written to stderr, but it's not very convenient.

This change could potentially break scripts using notionflux, but I find it quite unlikely.

Motivation for this change is to improve https://github.com/olejorgenb/notion-wm (yeah, I should really rename the repo)


raboof:
Cool! Better facilities for error handling are always useful :).

PR #43, opened 2017-04-11
Contrib: Glue code to use rofi as a menu replacement/alternative

olejorgenb:
[Rofi](https://github.com/DaveDavenport/rofi)

Adds `rofi.menu` - a `mod_menu.menu` compatible (mostly at least) replacement/alternative. 

An advantage over notion's menus is that it support fuzzier matching. (each word in the query is matched independently against the menu content making it order-independent for instance. The menu is also interactively filtered down.

Rofi support a built-in mode for switching windows which works with notion (when net_client_list is loaded), but the windows isn't listed in MRU order, and the rofi's built in workspace menu doesn't work IIRC.

Atm only regular menus is supported, but I don't think it would be very hard to support context menus too.

The mechanism used for passing data is a bit messy (suggestions welcome) but it works. 

Coauthored with @hedning 

2017-04-11-063312_1366x768_scrot

### DIFF
--- a/contrib/scripts/notion_rofi.sh
+++ b/contrib/scripts/notion_rofi.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env zsh
+#
+# Example:
+# rofi -show notion -modi "notion:'./notion_rofi.sh focuslist'" -scroll-method 1
+
+function escape {
+    sed "s/'/\\'/g"
+}
+
+MENU_NAME=$1
+
+if [[ -z $2 ]]; then
+    # No argument -> generate menu entries
+
+    notionflux -e "return rofi.menu_list('$MENU_NAME')"  \
+    | sed -e 's/\\"/"/g' -e 's/^"//' -e 's/\\$//g' -e '$d'
+else
+    # Entry selected
+
+    ENTRY=$(echo $2 | escape)
+    notionflux -e "rofi.menu_action('$MENU_NAME', '$ENTRY')" > /dev/null
+fi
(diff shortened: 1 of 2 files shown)


contrib/scripts/notion_rofi.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env zsh
+#
+# Example:
+# rofi -show notion -modi "notion:'./notion_rofi.sh focuslist'" -scroll-method 1
+
+function escape {
+    sed "s/'/\\\'/g"
```
raboof: Shouldn't existing `\` characters be somehow escaped, too?

raboof:
I'm not sure I completely understand how this works (and whether it makes sense to integrate it into Notion at all - couldn't it be completely separate?), but as it's in contrib I think it's fair game.

Thanks for the contribution!

contrib/scripts/notion_rofi.sh
olejorgenb: Yeah, you're right :)

It's not possible to activate a menu entry containing `\'`

olejorgenb:
Yeah, it's a bit messy, but I think it's kinda in the nature of how rofi and notionflux work.

olejorgenb:
> and whether it makes sense to integrate it into Notion at all - couldn't it be completely separate?

Some of it can be completely separate. "Go to window" is a built-in function in rofi and works when `net_client_list` is loaded into notion. But for stuff like workspaces, etc. I don't see how it's possible to not involve notion. (the list of window titles in the built-in rofi mode isn't MRU sorted either)

PR #46, opened 2017-06-15
Updated XFT support against master

ideasman42:
Update XFT patch for master, submitting as a PR.

Example font usage in `look` file:

    font = "xft:Cantarell:pixelsize=12:weight=bold",

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ matrix:
   exclude:
     - compiler: clang
       env: C99_SOURCE=""
-before_install: sudo apt-get install lua5.2 liblua5.2-dev gettext
+before_install: sudo apt-get install lua5.2 liblua5.2-dev gettext libxft-dev
 script:
   - make
   - make test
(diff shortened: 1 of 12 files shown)


raboof:
As discussed on IRC:
* I really love having access to XFT, especially on the increasingly common HiDPI screens it doesn't have a lot of downsides and greatly improves font availability.
* I'm not a huge fan of all the `#ifdef`ing, so in theory I'd prefer just moving everything over to XFT and stop supporting 'raw' X11 fonts. This will be unacceptable for our users on constrained or exotic systems, though, so `#ifdef` is our only option here. This is why I kept XFT support in the `xft` branch until now.
* If we can configure Travis to always build both with and without XFT that would take away some of this pain.
* For maintainers: they can enable XFT without impacting existing users, as XFT fonts are configured with the `xft:` prefix.

Could you update the travis config to build both with and without XFT?

raboof:
LOL it seems travis was already set up to build both with and without XFT, but didn't have libxft installed yet.

Thanks for the PR!

PR #49, opened 2017-08-10
Support for coexisting xft and bitmap fonts

ideasman42:
This commit makes it possible to have xft+bitmap fonts (both or either).

Currently bitmap fonts are always enabled, (to use when "xft:" prefix isnt found).

Note that we could disable bitmap support entirely by not defining `HAVE_X11_BMF`, In the current makefiles this isnt done.

### DIFF
--- a/de/colour.h
+++ b/de/colour.h
@@ -26,6 +26,11 @@ typedef XftColor DEColour;
 typedef unsigned long DEColour;
 #endif /* HAVE_X11_XFT */
 
+#ifdef HAVE_X11_XFT
+#  define PIXEL(x) (x).pixel
+#else /* HAVE_X11_XFT */
+#  define PIXEL(x) x
+#endif /* HAVE_X11_XFT */
 
 DECLSTRUCT(DEColourGroup){
     GrStyleSpec spec;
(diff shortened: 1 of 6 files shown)


de/colour.h
```diff
@@ -26,6 +26,11 @@ typedef XftColor DEColour;
 typedef unsigned long DEColour;
 #endif /* HAVE_X11_XFT */
 
+#ifdef HAVE_X11_XFT
```
ideasman42: Could use more verbose name since the define is moved into a header. eg `PIXEL_COLOR`

de/font.c
```diff
@@ -133,7 +136,11 @@ DEFont *de_load_font(const char *fontname)
     if(strncmp(fontname, "xft:", 4)==0){
         font=XftFontOpenName(ioncore_g.dpy, DefaultScreen(ioncore_g.dpy), fontname+4);
     }else{
+#ifdef HAVE_X11_BMF
```
ideasman42: Logic here is a bit messier than I'd like - but means we can build with/without bitmap fonts.

system-autodetect.mk
```diff
@@ -112,6 +112,9 @@ ifeq ($(USE_XFT),1)
     X11_INCLUDES += `pkg-config xft --cflags`
     X11_LIBS += `pkg-config xft --libs`
     DEFINES += -DHAVE_X11_XFT
+    DEFINES += -DHAVE_X11_BMF
```
ideasman42: This could be optional, for now use bitmap fonts with XFT.

PR #52, opened 2017-09-10
Display dialog windows as floating

ideasman42:
Currently dialogs with gtk2 flicker (gimps color picker flickers in tiled mode).
*(I removed windowing rules for gimp in my configuration since I use it in windowed mode).*

Since the window has the dialog property it can be set to floating.

Note that I wasn't sure where to assign the `float` property exactly, it's possible this should become a flag, so creating the property table isnt needed.

An alternative could be to set them transient, but this would mean it needs a parent window, which might not always be available, so I think floating is OK.

Even if this patch isn't used as-is, it should be able to be modified to make it acceptable.

----

Possible changes:

- Use a flag to tell the window to be floating instead of adding a table.
- Consider dialog windows *transient* (use notions transient rules here), accept that it's possible but unlikely for windows without a logical parent to have the dialog property. Note that I didn't see a good way to detect this, besides using the `_NET_WM_PID` to detect the parent window.

  I looked into this briefly and while it seems possible, ultimately I think its too error prone (all checks for `XGetTransientForHint` would need to do netwm PID lookups).
- Make this a configuration option so dialog behavior can be set to: [NONE, FLOAT, TRANSIENT].
- Expose the window-type property to the lua API and make the user defined callback responsible choosing what to do in the case of dialog windows.

  This may be the best option, however buggy-by-default is bad, so would suggest to also use a manage hook in the default configuration that handles dialogs.


ioncore/netwm.c
```diff
@@ -270,9 +274,33 @@ char **netwm_get_name(WClientWin *cwin)
     return xwindow_get_text_property(cwin->win, atom_net_wm_name, NULL);
 }
 
+/*}}}*/
+
+
+/*{{{ _NET_WM_WINDOW_TYPE */
+
+static bool netwm_window_type_is_dialog(WClientWin *cwin)
```
raboof: perhaps we should split out a `static Atom netwm_window_type(WClientWin *cwin)`?

ioncore/netwm.h
```diff
@@ -30,6 +30,7 @@ extern char **netwm_get_name(WClientWin *cwin);
 extern void netwm_handle_client_message(const XClientMessageEvent *ev);
 extern bool netwm_handle_property(WClientWin *cwin, const XPropertyEvent *ev);
 
+extern void netwm_float_dialog(WClientWin *cwin);
```
raboof: Hmm, I'm not sure I like exposing this as a side-effecting thing... I wonder if we could expose `netwm_window_type` and  `atom_net_wm_window_type_dialog` and then keep the logic inside `clientwin.c`?

ideasman42: Sounds better

ioncore/netwm.c
ideasman42: agree

ideasman42:
Updated the patch to use a flag instead of setting the property (the flag is only use when the property is unset)

ioncore/netwm.c
```diff
@@ -270,9 +274,26 @@ char **netwm_get_name(WClientWin *cwin)
...
 
+/*}}}*/
+
+
+/*{{{ _NET_WM_WINDOW_TYPE */
+
+Atom netwm_window_type(WClientWin *cwin)
+{
+    Atom p;
+    if(xwindow_get_atom_property(cwin->win, atom_net_wm_window_type, &p)){
```
raboof: I guess we could even have `xwindow_get_atom_property` return `Atom` - we might not even need `netwm_window_type` then. But this is OK :+1:

ioncore/region.h
```diff
@@ -26,7 +26,8 @@
 #define REGION_SKIP_FOCUS           0x0200
 #define REGION_CWINS_BEING_RESCUED  0x0400
 #define REGION_PLEASE_WARP          0x0800
-#define REGION_BINDING_UPDATE_SCHEDULED 0x1000
+#define REGION_PLEASE_FLOAT         0x1000
+#define REGION_BINDING_UPDATE_SCHEDULED 0x2000
```
raboof: I guess this is OK as we're not a library

ioncore/netwm.c
ideasman42: Wasn't sure about this because its possible the value Xlib stores is None in some cases.

PR #55, opened 2017-09-12
Support for window placement under mouse pointer

ideasman42:
Option to place new floating windows under the cursor (clamped by the workspace bounds).

Most window managers have this as an option, it's especially handy for dialog windows (may even be a good default for dialogs).

### DIFF
--- a/etc/cfg_notion.lua
+++ b/etc/cfg_notion.lua
@@ -66,6 +66,10 @@ ioncore.set{
     -- Auto-unsqueeze transients/menus/queries.
     --unsqueeze=true,
 
+    -- Float window placement method:
+    -- one of 'udlr' (up down left right), 'lrud' (left right up down), 'pointer' or 'random'.
+    --float_placement_method="udlr"
+
     -- Float non-transient dialog type windows.
     -- Can be useful for applications that try to place their own dialog windows.
     --window_dialog_float=true,
--- a/ioncore/float-placement.c
+++ b/ioncore/float-placement.c
@@ -9,6 +9,8 @@
 #include <string.h>
 
 #include "common.h"
+#include "rootwin.h"
+#include "xwindow.h"
 #include "group.h"
 #include "float-placement.h"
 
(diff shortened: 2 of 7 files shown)


etc/cfg_notion.lua
```diff
@@ -66,6 +66,10 @@ ioncore.set{
     -- Auto-unsqueeze transients/menus/queries.
     --unsqueeze=true,
 
+    -- Float window placement method:
+    -- one of 'udlr', 'lrud', 'pointer' or 'random'
+    --float_placement_method="udlr"
```
raboof: I know it's already in the reference docs, but perhaps document what 'udlr' means here again?

ioncore/float-placement.h
```diff
@@ -14,7 +14,7 @@
 
 
 typedef enum{
-    PLACEMENT_LRUD, PLACEMENT_UDLR, PLACEMENT_RANDOM
+    PLACEMENT_LRUD, PLACEMENT_UDLR, PLACEMENT_POINTER, PLACEMENT_RANDOM,
```
raboof: Put each option on a new line while we're at it?

ioncore/float-placement.c
```diff
@@ -167,10 +169,28 @@ static bool tiling_placement(WGroup *ws, uint level, WRectangle *g)
...
+}
+
 
 void group_calc_placement(WGroup *ws, uint level, WRectangle *geom)
 {
-    if(ioncore_placement_method!=PLACEMENT_RANDOM){
+    if(ioncore_placement_method==PLACEMENT_POINTER){
+        if(pointer_placement(ws, geom))
+            return;
+    }else if(ioncore_placement_method!=PLACEMENT_RANDOM){
```
raboof: Hmm, so it first tries to place according to POINTER/TILING depending on configuration, then when either of those fail falls back to RANDOM? Kind of weird but OK.

ioncore/rectangle.c
```diff
@@ -23,6 +23,24 @@ void rectangle_constrain(WRectangle *g, const WRectangle *bounds)
...
+    }
+
+    if(g->h>bounds->h){
+        g->y=(bounds->y+(bounds->h/2))-(g->h/2);
+    }else if(g->y<bounds->y){
+        g->y=bounds->y;
+    }else if(g->y+g->h>bounds->y+bounds->h){
+        g->y=(bounds->y+bounds->h)-g->h;
+    }
+}
```
raboof: This would typically be the kind of logic that could use a couple of simple unit tests ;) - but i gues we don't really have the infractructure for that in place properly.

ioncore/group-ws.c
```diff
@@ -56,12 +58,17 @@ void ioncore_groupws_set(ExtlTab tab)
...
-                          ? "lrud"
-                          : "random")));
+    const char *method;
+    if(ioncore_placement_method==PLACEMENT_UDLR){
+        method="udlr";
+    }else if(ioncore_placement_method==PLACEMENT_LRUD){
+        method="lrud";
+    }else if(ioncore_placement_method==PLACEMENT_POINTER){
+        method="pointer";
+    }else{
```
raboof: Much nicer than the 'staircase' above :). Perhaps log a warning when the method matches none of the known constants?

etc/cfg_notion.lua
ideasman42: done

ioncore/float-placement.h
ideasman42: done

ioncore/float-placement.c
ideasman42: Agree its strange,
Personally I think under-pointer may be better fallback.

I just kept functionality as-is.

ioncore/rectangle.c
ideasman42: normally logic like this isn't breaking between releases.

ioncore/group-ws.c
ideasman42: done

ideasman42:
Hopefully addressed all requests

ioncore/rectangle.c
raboof: Of course tests are also useful to validate the code works in the first place, and to document how it is intended to work. But no biggie here :)

PR #56, opened 2017-09-17
Makefiles: I make sure to invoke $(shell) only once per run

dkogan:
In a few places we used $(shell) in a deferred variable, meaning that the shell
would be invoked each time the expression is evaluated, which is a waste of
resources

### DIFF
--- a/build/lua-detect.mk
+++ b/build/lua-detect.mk
@@ -23,7 +23,9 @@ ifeq ($(LUA_PKG),)
     $(error Could not find $(or $(LUA_VERSION),any) lua version. (Did you install the -dev package?))
 endif
 
-LUA_VERSION ?= $(or $(shell $(PKG_CONFIG) --variable=V $(LUA_PKG)),5.0)
+ifeq ($(LUA_VERSION),)
+LUA_VERSION := $(or $(shell $(PKG_CONFIG) --variable=V $(LUA_PKG)),5.0)
+endif
 
 # prior to 5.1 the lib didn't include version in name.
 LUA_SUFFIX := $(if $(findstring $(LUA_VERSION),5.0),,$(LUA_VERSION))
--- a/mod_xrandr/Makefile
+++ b/mod_xrandr/Makefile
@@ -9,13 +9,17 @@ include $(TOPDIR)/build/system-inc.mk
 
 ######################################
 
-INCLUDES += $(LIBTU_INCLUDES) $(LIBEXTL_INCLUDES) $(X11_INCLUDES) -I$(TOPDIR) $(shell pkg-config --cflags-only-I xrandr)
-CFLAGS += $(XOPEN_SOURCE) $(C99_SOURCE) $(shell $(PKG_CONFIG) --cflags-only-other xrandr)
+XRANDR_CFLAGS_ONLYI     := $(shell $(PKG_CONFIG) --cflags-only-I xrandr)
+XRANDR_CFLAGS_ONLYOTHER := $(shell $(PKG_CONFIG) --cflags-only-other xrandr)
+XRANDR_LDLIBS           := $(shell $(PKG_CONFIG) --libs xrandr)
+
+INCLUDES += $(LIBTU_INCLUDES) $(LIBEXTL_INCLUDES) $(X11_INCLUDES) -I$(TOPDIR) $(XRANDR_CFLAGS_ONLYI)
+CFLAGS += $(XOPEN_SOURCE) $(C99_SOURCE) $(XRANDR_CFLAGS_ONLYOTHER)
 
 SOURCES=mod_xrandr.c
 
 MAKE_EXPORTS=mod_xrandr
-LIBS = $(X11_LIBS) $(shell $(PKG_CONFIG) --libs xrandr)
+LIBS = $(X11_LIBS) $(XRANDR_LDLIBS)
 MODULE=mod_xrandr
 MODULE_STUB=mod_xrandr.lua
 ETC=cfg_xrandr.lua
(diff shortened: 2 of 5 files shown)


build/lua-detect.mk
```diff
@@ -23,7 +23,9 @@ ifeq ($(LUA_PKG),)
     $(error Could not find $(or $(LUA_VERSION),any) lua version. (Did you install the -dev package?))
 endif
 
-LUA_VERSION ?= $(or $(shell $(PKG_CONFIG) --variable=V $(LUA_PKG)),5.0)
+ifeq ($(LUA_VERSION),)
+LUA_VERSION := $(or $(shell $(PKG_CONFIG) --variable=V $(LUA_PKG)),5.0)
+endif
```
raboof: What's the difference here?

mod_xrandr/Makefile
```diff
@@ -9,13 +9,17 @@ include $(TOPDIR)/build/system-inc.mk
...
 ######################################
 
-INCLUDES += $(LIBTU_INCLUDES) $(LIBEXTL_INCLUDES) $(X11_INCLUDES) -I$(TOPDIR) $(shell pkg-config --cflags-only-I xrandr)
-CFLAGS += $(XOPEN_SOURCE) $(C99_SOURCE) $(shell $(PKG_CONFIG) --cflags-only-other xrandr)
+XRANDR_CFLAGS_ONLYI     := $(shell $(PKG_CONFIG) --cflags-only-I xrandr)
+XRANDR_CFLAGS_ONLYOTHER := $(shell $(PKG_CONFIG) --cflags-only-other xrandr)
+XRANDR_LDLIBS           := $(shell $(PKG_CONFIG) --libs xrandr)
+
+INCLUDES += $(LIBTU_INCLUDES) $(LIBEXTL_INCLUDES) $(X11_INCLUDES) -I$(TOPDIR) $(XRANDR_CFLAGS_ONLYI)
+CFLAGS += $(XOPEN_SOURCE) $(C99_SOURCE) $(XRANDR_CFLAGS_ONLYOTHER)
```
raboof: :+1: nice, makes it more readable as well

build/lua-detect.mk
dkogan: The old version creates a deferred variable (evaluated each time it is expanded); the new version creates an immediate variable (evaluated once; now)

dkogan: I should note this pull request is 100% changing variables with $(shell) from deferred variables to immediate ones. The effect is that each $(shell) is invoked just once.

raboof: In this case we seem to be trading readability for speed - but when I try it it doesn't appear to produce any noticeable speed improvement? Or do you know a situation where it is evident?

dkogan:
Arnout Engelen <notifications@github.com> writes:

> raboof commented on this pull request.
>
>> @@ -23,7 +23,9 @@ ifeq ($(LUA_PKG),)
>      $(error Could not find $(or $(LUA_VERSION),any) lua version. (Did you install the -dev package?))
>  endif
>
> -LUA_VERSION ?= $(or $(shell $(PKG_CONFIG) --variable=V $(LUA_PKG)),5.0)
> +ifeq ($(LUA_VERSION),)
> +LUA_VERSION := $(or $(shell $(PKG_CONFIG) --variable=V $(LUA_PKG)),5.0)
> +endif
>
> In this case we seem to be trading readability for speed - but when I
> try it it doesn't appear to produce any noticeable speed improvement?
> Or do you know a situation where it is evident?

As a general rule, it's poor form to ever put $(shell) into a deferred
variable. I don't know where $(LUA_VERSION) is used. Maybe it's used
inside some build flags, so we invoke an extra shell for every .o we
compile. Or maybe it's not used that way YET, but somebody will use it
that way in the future. It shouldn't matter. If you don't have a
$(shell) in a deferred variable, this cannot ever become a problem.

knixeur:
Adding some light to this change:

https://www.gnu.org/software/make/manual/html_node/Flavors.html#Flavors

https://www.gnu.org/software/make/manual/html_node/Setting.html#Setting
```
FOO ?= bar
```
is equal to
```
ifeq ($(origin FOO), undefined)
  FOO = bar
endif
```
Which is a a recursively expanded variable assignment.
The official examples uses $(shell) with simple expanded form
https://www.gnu.org/software/make/manual/html_node/Shell-Function.html

I guess there's not much optimization because with ?= once it's only expanded when it's not defined.

dkogan:
Guillermo Bonvehí <notifications@github.com> writes:

> Adding some light to this change:
>
> https://www.gnu.org/software/make/manual/html_node/Flavors.html#Flavors
>
> https://www.gnu.org/software/make/manual/html_node/Setting.html#Setting
> ```
> FOO ?= bar
> ```
> is equal to
> ```
> ifeq ($(origin FOO), undefined)
>   FOO = bar
> endif
> ```
> Which is a a recursively expanded variable assignment.
> The official examples uses $(shell) with simple expanded form
> https://www.gnu.org/software/make/manual/html_node/Shell-Function.html
>
> I guess there's not much optimization because with ?= once it's only expanded when it's not defined.

No. The big issue is that you get 'FOO = bar' and not 'FOO := bar'. So
each time $(FOO) is evaluated it invokes the shell. Look at this
Makefile:

```
ifeq ($(origin FOO), undefined)
  FOO := $(shell echo xxx > /dev/stderr)
endif

$(info $(FOO))
$(info $(FOO))
$(info $(FOO))
```

When you 'make' with this, it prints "xxx" once. But if you had =
instead of :=, it would print "xxx" 3 times.

knixeur:
You're right, I understood the command incorrectly, for reference I'm adding a ?= example
```
BAR ?= $(shell echo bar > /dev/stderr)
$(info $(BAR))
$(info $(BAR))
$(info $(BAR))
```
prints "bar" 3 times

So indeed changing to := would improve by not reevaluating as ?= is.

knixeur:
So these changes look good to be merged, it should speed things up (may not be noticeable in all situations) but also the recommended way to use shell according to make documentation.

PR #60, opened 2018-01-06
Fix bitmap font display

ideasman42:
Bitmap font's allocate this as needed, checking caused text not to display.

### DIFF
--- a/de/font.c
+++ b/de/font.c
@@ -501,7 +501,7 @@ void debrush_do_draw_string_default(
         }
 #endif
 #ifdef HAVE_X11_BMF
-        if (brush->d->font->fontset) {
+        {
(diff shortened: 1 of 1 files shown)


ideasman42:
For some reason this did not happen for me with the original patch. but is now needed.

de/font.c
```diff
@@ -501,7 +501,7 @@ void debrush_do_draw_string_default(
         }
 #endif
 #ifdef HAVE_X11_BMF
-        if (brush->d->font->fontset) {
+        {
```
raboof: Hmm, wouldn't this cause an XFT font to be drawn twice in the case that both `HAVE_X11_XFT` and `HAVE_XFT_BMF` are true?

ideasman42: No, XFT returns if it's found.

raboof: aah of course 👍

PR #70, opened 2019-03-16
Xutf8TextListToTextProperty needs a NULL terminated list of pointers.

0-wiz-0:


raboof:
Are you sure? the manpage suggests the `list` must contain null-terminated strings, but not that it must be null-terminated itself. I would expect it just reads `count` elements from the list?

0-wiz-0:
No, I'm not sure. I don't trust the system on which I did the bugtracking, and on second reading I understand the man page your way.

PR #76, opened 2019-04-17
statusd_inetaddr2: Fixed reading using get_output_tool

knixeur:
I guess I tested an old version before uploading latest change, sorry
about that.

### DIFF
--- a/contrib/statusd/statusd_inetaddr2/statusd_inetaddr2.lua
+++ b/contrib/statusd/statusd_inetaddr2/statusd_inetaddr2.lua
@@ -50,7 +50,7 @@ local function get_tool_output(command)
     if not f then
         return nil
     end
-    lines = f:read("a")
+    lines = f:read("*a")
     f:close()
     return lines
 end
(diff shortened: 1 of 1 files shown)


contrib/statusd/statusd_inetaddr2/statusd_inetaddr2.lua
```diff
@@ -124,8 +124,6 @@ local function update_inetaddr()
     inetaddr2_timer:set(settings.interval, update_inetaddr)
 end
 
-inetaddr2_timer:set(100, update_inetaddr)
-
 if settings.mode == "ip" then
```
wilhelmy: You might want to change this to local, perhaps like
```lua
local ip_paths = settings.mode == "ip"
  and  {"/sbin/ip", "/usr/sbin/ip", "/bin/ip", "/usr/bin/ip"}
  or   {"/sbin/ifconfig", "/usr/sbin/ifconfig", "/bin/ifconfig", "/usr/bin/ifconfig"}
```

or just declaring `local ip_paths` before the conditional, lua makes everything global by default.

knixeur: Thanks for the review, I think I'll declare local ip_paths = {} before, that way if settings.mode is not correctly configured the assert will fail.

PR #77, opened 2019-04-17
Fix using a non xft font with xft compiled

knixeur:
Due to #ifndef XFT, which is defined in this case, a step to set
the style was beign skipped when using a BMF font.
Free BMF fonts even when XFT is defined.
Simplified return value in defont_get_text_width.

### DIFF
--- a/de/font.c
+++ b/de/font.c
@@ -221,13 +221,12 @@ bool de_set_font_for_style(DEStyle *style, DEFont *font)
     style->font=font;
     font->refcount++;
 
-#ifndef HAVE_X11_XFT
+#ifdef HAVE_X11_BMF
     if(style->font->fontstruct!=NULL){
         XSetFont(ioncore_g.dpy, style->normal_gc,
                  style->font->fontstruct->fid);
     }
-
-#endif /* ! HAVE_X11_XFT */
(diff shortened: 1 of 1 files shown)


knixeur:
With issue
https://pasteboard.co/Iay2C8o.gif
Fixed
https://pasteboard.co/Iay2iqQ.gif

I cannot upload gifs here :/

de/font.c
```diff
@@ -373,9 +373,7 @@ uint defont_get_text_width(DEFont *font, const char *text, uint len)
         }
     }
 #endif /* HAVE_X11_BMF */
-    else{
-        return 0;
-    }
+    return 0;
```
raboof: I'm not sure I like this change: like this, the compiler no longer checks that all the other branches above have their own `return` statement

knixeur: You're right about that but if none of the flags XFT or BMF are defined, it's actually invalid code (there's no if). Also, the way it's structured, it needs to have a return because it tries first by font, then fontset, then fontstruct otherwise it will bail to 0. If any of those don't return, another branch will may execute and yield an incorrect result.

raboof: Ah yes makes sense :+1:

PR #79, opened 2019-04-20
WIP: Reactivate tests

knixeur:
I couldn't make integration tests work. I converted the code to use new Lua posix library but then had troubles with Xdummy so I gave up on that part.
I added some tests for libextl to test @olejorgenb 's boolean support
Also made a simple test using existing tester.c in libtu
The existing tests are really simple but could be the starting point to add more in the future if/when bugs are found. I was thinking this would help with warning cleanup branch but the tests are really simple and integration ones don't work.

This would also address #38 as it includes his addition

Edit: typo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ utils/ion-statusd/ion-statusd
 utils/ion-completefile/ion-completefile
 cscope.*
 test/integration/libXinerama.so.1.0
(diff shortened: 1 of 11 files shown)


knixeur:
Not really, it works, I just wanted to wait a few days in case someone had an improvement to make.

PR #83, opened 2019-05-01
Add Contributing section to README

raboof:


knixeur:
http://notion.sourceforge.net/notionnotes/ and http://notion.sourceforge.net/notionconf/ should be moved to notionwm.net

raboof:
I agree, but they haven't yet, and I don't think that needs to block this
PR. If you can ticketify or even PR the move of those 2 resources that
would be great!

PR #85, opened 2019-05-02
[WIP] Reactivate integration tests

knixeur:
Added check for lua-posix module, it can be installed with distribution's package manager or lua rocks

- [ ] Explain on how to install Lua Rocks

- [ ] Autodetect (which?) Xorg binary directory

- [ ] Explain that xorg dummy driver is required for integration testing


wilhelmy:
This is marked as WIP, what's missing here? I think integration tests might be a good idea if it isn't a lot of effort to finish this off.

knixeur:
I think it's ready, it works like it did before. I think I need someone to test it too and then finish the items in the checklist (luarocks + docs)

test/integration/runtests.lua
```diff
@@ -1,5 +1,13 @@
-# on Debian in lua-posix
-require "posix"
+--# on Debian in lua-posix
+local basedir = "../../"
+
+print("[TESTING] Integration")
+
+local success, posix = pcall(require, "posix")
+if not success then
+    print("[SKIP] posix module not found")
```
raboof: Shouldn't we fail in this case? I guess that means adding lua-posix to the .travis.yml

raboof:
very cool to replace that XDummy script with a configuration :+1: great work here

test/integration/runtests.lua
knixeur: Mmm I think we should either fail but run integration tests only when running make test or skip if we do run integration tests during a normal make.
I think the best approach is to run integration tests only when make test is run explicitly so lua-posix is not needed when running a common make.

raboof: I agree - I thought we already didn't run the tests on normal 'make' ;)

knixeur: I thought we did, now it fails on error

raboof:
I guess on travis we might use https://docs.travis-ci.com/user/gui-and-headless-browsers/#using-xvfb-to-run-tests-that-require-a-gui - but then perhaps we don't need to start our own X11 server in the script?

test/integration/xrandr/01_remove_display.lua
```diff
@@ -22,8 +67,8 @@ print('Updating layout when there is 2 screens present')
 -- remove one, and then leave checking that one was removed up to the next test.
 mod_xrandr.screenlayoutupdated()
 
-if notioncore.find_screen_id(0):mx_count() ~= 1 then
-  return "After updating screen 0 should have 1 workspaces instead of " ..
+if notioncore.find_screen_id(0):mx_count() ~= 2 then
```
raboof: so updating the screen layout, with 2 screens connected, 'spontaneously' another workspace appears? That doesn't look right - might be a problem with the test setup of course, I'll try and see if I can explain this one

test/integration/xrandr/02_removed_display.lua
```diff
@@ -1,11 +1,56 @@
...
+        os.execute("cp xrandr/fakexinerama-2monitors xrandr/.fakexinerama")
+    end
+end
+-- /TEMP
+
 mod_xrandr.screenlayoutupdated()
 
-if notioncore.find_screen_id(1) then
-  return "New number of screens should be 1, found ", notioncore.find_screen_id(1):name()
+if notioncore.find_screen_id(1) ~= nil then
```
raboof: :+1:

test/integration/xrandr/02_removed_display.lua
```diff
@@ -1,11 +1,56 @@
...
+  return "New number of screens should be 1, found " .. notioncore.find_screen_id(1):name()
 end
 
-if notioncore.find_screen_id(0):mx_count() ~= 2 then
-  return "Remaining screen should have 2 workspaces instead of " .. notioncore.find_screen_id(0):mx_count() ..
+if notioncore.find_screen_id(0):mx_count() ~= 3 then
```
raboof: Seems surprising - but perhaps a side-effect of the previous test?

test/integration/xrandr/03_new_screen_has_ws.lua
```diff
@@ -13,8 +58,11 @@ end
 if notioncore.find_screen_id(0):mx_nth(0):name() ~= 'WGroupWS' then
   return "First workspace not correctly returned to first screen"
 end
-if notioncore.find_screen_id(1):mx_nth(0):name() ~= 'WGroupWS<1>' then
-  return "Second workspace not correctly returned to second screen"
+if notioncore.find_screen_id(0):mx_nth(1):name() ~= 'WGroupWS<1>' then
+  return "Second workspace correctly kept on first screen"
```
raboof: I think the intent here was the test would return either `"ok"` or an error

test/integration/xrandr/02_removed_display.lua
knixeur: Could be, I should do some more testing here

test/integration/xrandr/03_new_screen_has_ws.lua
knixeur: I guess something like "Second workspace not kept in first screen" if that's the desired behaviour

test/integration/xrandr/02_removed_display.lua
knixeur: Test 02 of xinerama suffers from the same issue, if test 01 is not ran it fails. I wonder when this started to occur since I didn't modify that one.

wilhelmy:
So... do we merge this?

knixeur:
I agree with @raboof here, I think we should really understand the tests before merging this.
I don't know if the tests were wrong before I started working on this because I couldn't run them (iirc..). It might be possible to run the tests on master and see the results to compare against this.
If someone wants to give it a try I'll glad, I'm not finding free time lately to finish this.

test/integration/xrandr/01_remove_display.lua
wilhelmy: I think I've seen this on my setup, if I attach a second screen, it comes with two workspaces.

PR #86, opened 2019-05-03
Use NOTION_RELEASE environment value if defined as version and sh compat

knixeur:
~Use technique suggested in https://stackoverflow.com/q/7292584
to check git can be executed~

Edit: I've updated the PR, we talked in IRC and an environment value is a better fit

### DIFF
--- a/version.sh
+++ b/version.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 # Determines the currently checked-out version of Notion.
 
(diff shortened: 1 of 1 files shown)


knixeur:
You're right, I'm swallowing the version it should go to the else condition too.

raboof:
The 'else' expects you to be in a directory that follows the github tag zip download naming conventions - if there is a .git directory present then it's unlikely the directory name follows that convention...

version.sh
```diff
@@ -20,10 +20,15 @@
 
 set -e
 
+if [ ! -z "$NOTION_RELEASE" ]; then
+    echo "$NOTION_RELEASE"
+    exit
```
raboof: Can we use `elif` instead?

version.sh
```diff
@@ -20,10 +20,15 @@
...
+fi
+
 if [ -e ".git" ]; then
   # Git:
   echo -n `git describe`
-  if [[ -n $(git status -s) ]]; then
+  if [ ! -z "$(echo -n `git status -s`)" ]; then
```
raboof: Nice :+1:

version.sh
knixeur: Sure, done :)

Issue #84, opened 2019-05-01
Reproducible Builds


raboof:
A useful advantage of Open Source software is the fact that it is possible to check the sources for unwanted code.

When software is distributed in binary form, it does not go without saying that the distributed binary indeed corresponds to the sources. An elegant way to increase confidence that the build/distribution pipeline has not been compromised is to independently build the application on different machines, and check that the resulting binaries match. This does require that we make sure there are no 'accidental' non-determinisms in our build process. This is what we mean by making the build 'reproducible'.

'Officially', Notion is only distributed in source form, side-stepping this issue entirely. If you like to create binary packages for Notion, however, we warmly recommend you build them in a 'reproducible' way. It is entirely possible that this requires improvements to the Notion build setup, and we welcome any contributions that help make it more reproducible.

A useful resource around this topic is https://reproducible-builds.org

Subtasks:

* [x] exports.c generation is not deterministic
* [x] Test with reprotest

raboof:
This was manually verified and looks OK, but it would be good to check for regressions automatically - filed #106 for that.

PR #100, opened 2019-05-11
Produce libextl exports in deterministic order

raboof:
Part of #84


raboof:
Actually `reprotest "make clean && make" notion/notion` suggests we're already reproducible with this change. Surprising!

PR #120, opened 2019-06-20
Add empty release drafter config

raboof:


raboof:
consensus in #121 seems to be that we want to tag-to-release, so let's give the drafter a try

PR #122, opened 2019-06-21
etc/Makefile: honour $(DESTDIR) like every other Makefile

wilhelmy:


wilhelmy:
Previously, `make install DESTDIR=testdir PREFIX=foo` would fail.

raboof:
@knixeur feel free to merge when you're confident about a review :smile:

PR #137, opened 2019-06-27
Support submaps (binding and entering) when no kpress binding is defined

### LINKED ISSUE #128: Submaps are not entered if there is not also a kpress handler for that combination
Introduced in #105 
Found by @wilhelmy

raboof:
Fixes #128

### DIFF
--- a/ioncore/key.c
+++ b/ioncore/key.c
@@ -244,12 +244,9 @@ static bool do_key(WRegion *reg, XKeyEvent *ev)
     }
 
     has_submap = submap_defined(oreg, ev);
-    if(has_submap){
-        if(add_sub(oreg, ev->keycode, ev->state))
-            grab_needed = TRUE;
-        else
-            clear_subs(oreg);
-    }else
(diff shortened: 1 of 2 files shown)


raboof:
Wait, something is not right here yet: with this change, `META+Button3` shows the indicator for 'move' rather than for 'resize' (even though it correctly resizes).

ioncore/key.c
```diff
@@ -249,22 +249,20 @@ static bool do_key(WRegion *reg, XKeyEvent *ev)
             grab_needed = TRUE;
         else
             clear_subs(oreg);
-    }
+    }else
```
knixeur: Maybe this can be condensed to avoid double clear_subs calls.
Ie:
```
if(has_submap && add_sub(oreg, ev->keycode, ev->state)) {
    grab_needed = TRUE;
} else {
    clear_subs(oreg);
}
```

raboof: Sounds good. Are you familiar with the `Insert a suggestion <ctrl-g>` feature of github? :)

PR #140, opened 2019-06-29
Add migration guide

raboof:
refs #113

### DIFF
--- a/migration.md
+++ b/migration.md
@@ -1,0 +1,64 @@
+---
+layout: default
+---
+
+# Migration
+
+The next major release of Notion will be Notion 4. This release will include a
+number of changes that are either not backwards compatible, or change the
+(default) behavior.
+
+## Custom styles
+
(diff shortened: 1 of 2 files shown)


wilhelmy:
If I'm not mistaken, we still need to link to this document from the main page.

raboof:
Good point. I didn't because Notion 4 isn't released yet, but indeed it makes sense to make this easy to find for people building from master. Added.

migration.md
```diff
@@ -0,0 +1,107 @@
...
+```
+becomes
+```lua
+function foo(a, b, ...)
+    print(...)
+end
+```
+and capturing `...` into a table can be achieved with `local arg={...}`.
+
+Notion 4 also requires a C compiler capable of compiling C99.
```
knixeur: I'd add a subtitle or something to make this more visible. Maybe Lua/C Compiler subheaders

PR #142, opened 2019-07-01
Use an embedded dock by default, instead of mod_statusbar

### LINKED ISSUE #115: Make mod_dock the default
Replacing mod_statusbar. Clarify in migration guide (#113)

raboof:
Fixes #115


wilhelmy:
This means that after installing notion, the dock area will be empty, right?

raboof:
Yes, it will be invisible until you start a dock such as xfce4-panel

wilhelmy:
Right. My dock used to have a minimum size, which was honoured by the previous version of notion I was running and thus presented itself as a 64px high bar on the top of the screen. Ever since I upgraded to master, it is gone, though (I guess because it no longer supports minimum size hints).

raboof:
Sounds like something that would be reasonable to support, filed #149

Issue #112, opened 2019-06-18
'discover notion' interactive guide?


raboof:
It would be neat to have an 'interactive guide' that shows all keybindings, highlighting ones that you have already used, recommending which new ones you have yet to try.

wilhelmy:
Something like vimtutor? If so, sounds like a good idea.

PR #143, opened 2019-07-01
New keybindings

raboof:


etc/cfg_bindings.lua
```diff
@@ -0,0 +1,397 @@
...
+    kpress(META.."Caps_Lock", "WFrame.switch_next(_)"),
+    mclick(META.."Button4", "WFrame.switch_next(_)"),
+    mclick(META.."Button5", "WFrame.switch_prev(_)"),
+})
+
+-- Frames for transient windows ignore this bindmap
+
+defbindings("WFrame.toplevel", {
+    bdoc("Query for a client window to attach."),
+    kpress(META.."X", "mod_query.query_attachclient(_)"),
```
wilhelmy: I think "A" for "attach" was quite memorable

etc/cfg_bindings.lua
```diff
@@ -0,0 +1,397 @@
...
+        bdoc("Move current object within the frame left/right."),
+        kpress("comma", "WFrame.dec_index(_, _sub)", "_sub:non-nil"),
+        kpress("period", "WFrame.inc_index(_, _sub)", "_sub:non-nil"),
+
+        bdoc("Maximize the frame horizontally/vertically."),
+        kpress("H", "WFrame.maximize_horiz(_)"),
+        kpress("V", "WFrame.maximize_vert(_)"),
+
+        bdoc("Attach tagged objects to this frame."),
+        kpress("X", "ioncore.tagged_attach(_)"),
```
wilhelmy: Doesn't this conflict with META.."K" "X" from cfg_tiling.lua? ("unsplit_at")
Edit: Maybe notion should warn in case of duplicate bindings?

etc/cfg_bindings.lua
```diff
@@ -0,0 +1,397 @@
...
+
+defbindings("WFrame.toplevel", {
+    bdoc("Query for a client window to attach."),
+    kpress(META.."X", "mod_query.query_attachclient(_)"),
+
+    kpress(META.."A", "WFrame.switch_nth(_, 0)"),
+    kpress(META.."S", "WFrame.switch_nth(_, 1)"),
+    kpress(META.."D", "WFrame.switch_nth(_, 2)"),
+    kpress(META.."F", "WFrame.switch_nth(_, 3)"),
+
```
wilhelmy: How about META..[F1…F9] to change workspaces and META..[1…9] to change client windows (or the other way around because changing workspaces is probably used more often and thus should be closer to the home row)?

etc/cfg_bindings.lua
```diff
@@ -0,0 +1,397 @@
...
+    bdoc("Query for Lua code to execute."),
+    kpress(ALTMETA.."J", "mod_query.query_lua(_)"),
+
+    bdoc("Query for host to connect to with SSH."),
+    kpress(ALTMETA.."F4", "mod_query.query_ssh(_, ':ssh')"),
+
+    bdoc("Query for file to edit."),
+    kpress(ALTMETA.."F5",
+           "mod_query.query_editfile(_, 'run-mailcap --action=edit')"),
```
wilhelmy: Does `run-mailcap` even still work?

etc/cfg_bindings.lua
```diff
@@ -0,0 +1,397 @@
...
+        --kpress("K", "ioncore.goto_previous()"),
+
+        --bdoc("Go to first object on activity/urgency list."),
+        --kpress("I", "ioncore.goto_activity()"),
+
+        bdoc("Clear all tags."),
+        kpress("T", "ioncore.clear_tags()"),
+    }),
+
+    bdoc("Go to n:th screen on multihead setup."),
```
wilhelmy: intentional redundancy here?

etc/cfg_bindings.lua
```diff
@@ -0,0 +1,397 @@
...
+    kpress(META.."Shift+2", "ioncore.goto_nth_screen(1)"),
+    kpress(META.."F1", "ioncore.goto_nth_screen(1)"),
+    kpress(META.."Shift+3", "ioncore.goto_nth_screen(2)"),
+    kpress(META.."F2", "ioncore.goto_nth_screen(2)"),
+    kpress(META.."Shift+4", "ioncore.goto_nth_screen(2)"),
+    kpress(META.."F3", "ioncore.goto_nth_screen(2)"),
+
+    bdoc("Go to next/previous screen on multihead setup."),
+    kpress(ALTMETA.."Escape", "ioncore.goto_prev_screen()"),
+    kpress(META.."Escape", "ioncore.goto_next_screen()"),
```
wilhelmy: Bound to screen 0 in line 58?

etc/cfg_bindings.lua
```diff
@@ -0,0 +1,397 @@
...
+    bdoc("Display the main menu."),
+    kpress("F12", "mod_query.query_menu(_, _sub, 'mainmenu', 'Main menu:')"),
+    --kpress("F12", "mod_menu.menu(_, _sub, 'mainmenu', {big=true})"),
+    mpress("Button3", "mod_menu.pmenu(_, _sub, 'mainmenu')"),
+
+    bdoc("Display the window list menu."),
+    mpress("Button2", "mod_menu.pmenu(_, _sub, 'windowlist')"),
+
+    bdoc("Forward-circulate focus."),
+    -- '_chld' used here stands to for an actual child window that may not
```
wilhelmy: I think a link to the documentation for _chld, _sub and _ in the context of keybinding config would be really helpful here.

etc/cfg_bindings.lua
raboof: I like using 'A' for 'first tab', but indeed it might be good to move 'attach' elsewhere.

etc/cfg_bindings.lua
raboof: Yeah, there could be some conflicts that we need to fix.

Perhaps we should have a test that checks there are no duplicate bindings in the default configs. At run time the user is allowed to 'overwrite' bindings, so then a warning makes less sense.

etc/cfg_bindings.lua
raboof: The general concept is:
* META..[Esc,F1...F9] switches screens
* META..[`,1..9] switches workspace within a screen
* META..[Tab-Q] switches between tiles
, META..[CapsLocak,A-F] switches between tabs within a tile

etc/cfg_bindings.lua
raboof: Why wouldn't it? Though I agree it's a bit antiquated and perhaps we should switch to some freedesktop.org standard. Separate issue though.

etc/cfg_bindings.lua
raboof: Not really, I think we should remove the `META+Shift+n` ones.

etc/cfg_bindings.lua
raboof: Good one, those need to shift one place :+1:

etc/cfg_bindings.lua
wilhelmy: I've created an issue and I'm closing this.

etc/cfg_bindings.lua
wilhelmy: I usually have less than 10 screens but more than 5 tabs. F1…12 is a second range of consecutive number keys, I'd say it would be best to use them as such, with everything else on the keyboard being random letters, some of which we might want to use as a mnemonic (the way vim does, "o"=open, "a"=append etc.), which makes it complicated to allocate let's say the whole Q…P row for tab 1…n (besides, those keys are usually shuffled around pseudorandomly in a non-US-keymap).

After all, the keyboard layout is the first thing a new user sees about notion, so I feel it should be as memorable as possible in order to make it easy to get going (this is why I'm proposing "A" for "attach" again).

I'm also a bit conflicted about the META.."K" submap in general, it seems this was used as a place for everything that didn't fit anywhere else (like C-x in emacs), and does not have any meaning? I think it would make more sense to have submaps (if we're going to have any at all in the new keybinding setup) to operate on a type of object, like tabs, e.g. all tab keybindings in META.."T", all workspace keybindings in META.."W", all "go to window by property" (urgent or title) in META.."G" and so on?

Just my two cents, feel free to do something else entirely ;)

etc/cfg_bindings.lua
knixeur: I'd leave META+shift+n unless you find something more useful for them, in the keyboard at the office, Fx keys are far away from numbers.

etc/cfg_bindings.lua
wilhelmy: META.."P" for "put"?

etc/cfg_bindings.lua
wilhelmy: Maybe META.."`" could be the submap for whatever object is going to be on META..[0…9] (workspaces)
and META.."Escape" could be the submap for whatever object is going to be on META..[F1…F12] (screens or windows?)

etc/cfg_bindings.lua
wilhelmy: Some possible mnemonics: Put, Yank, Capture, Nick, Steal, Take

etc/cfg_bindings.lua
raboof: > Just my two cents, feel free to do something else entirely ;)

Thanks! I really value the discussion, even though I've been experimenting with this for a while so there are some concepts that I'm reluctant to change :)

> the keyboard layout is the first thing a new user sees about notion, so I feel it should be as memorable as possible in order to make it easy to get going

I agree (I wonder if some companion app like #112 could help here, too)

> everything else on the keyboard being random letters

Part of idea behind this binding is that all basic 'direct' keys are at the left hand (navigation, fullscreen, scratchpad, etc), and more rarely used keys (e.g. anything that opens a query) are at the right hand. I think this would greatly improve how memorable the keys are. I agree using mnemonics is a useful tool as well, but I think that should be a secondary concern.

> some of which we might want to use as a mnemonic (the way vim does, "o"=open, "a"=append etc.)

vim is a bit of an unfortunate example, with hjkl having special meaning there :)

> those keys are usually shuffled around pseudorandomly in a non-US-keymap

This is definitely a downside, but I think we should just leave it up to the community to come up with bindings for use with different keymaps.

> I'm also a bit conflicted about the META.."K" submap in general

I'm not a fan either. I wonder if we could get away with putting everything under `META` and `META+Shift` and drop `META+K` entirely.

etc/cfg_bindings.lua
raboof: That makes sense, let's leave them in for now.

etc/cfg_bindings.lua
raboof:
```
Maybe META.."`" could be the submap for whatever object is going to be on META..[0…9] (workspaces)
and META.."Escape" could be the submap for whatever object is going to be on META..[F1…F12] (screens or windows?)
```
For those I had in mind 'next' (so `META+Escape`=next screen, `META+\``=next ws, `META+Tab`=next frame, `META+CapsLock`=next tab)

etc/cfg_bindings.lua
raboof: I like those! Thx!

etc/cfg_bindings.lua
wilhelmy: TIL. I had a copy of all these cfg_*.lua files when I could have just changed the parts I don't like. Silly me.

raboof: it's even documented at http://notion.sourceforge.net/notionconf/node4.html#sec:bindings - but kind of hidden in the wall of text :laughing:

etc/cfg_bindings.lua
wilhelmy: > Thanks! I really value the discussion, even though I've been experimenting with this for a while so there are some concepts that I'm reluctant to change :)

I understand. I'm most likely not going to use this new setup (except for parts that I like more than what we had in my own modified version of the notion3 standard keymap) because I don't like changing my habits for no good reason, but I think new users should have it easy. If you think you have that figured out, I'm not going to cause you some bikeshed discussion about it.

> I agree (I wonder if some companion app like #112 could help here, too)

I think so. `vimtutor` helped me getting started with vim, at least. This would be very easy to integrate into notion on first start thanks to the lua interface. "Would you like to try the tutorial mode? (y/n)".

> Part of idea behind this binding is that all basic 'direct' keys are at the left hand (navigation, fullscreen, scratchpad, etc), and more rarely used keys (e.g. anything that opens a query) are at the right hand. I think this would greatly improve how memorable the keys are. I agree using mnemonics is a useful tool as well, but I think that should be a secondary concern.

I'm inclined to say that mnemonics and similarity to other tools (something that old-school *nix generally does rather badly compared to, say, OS X) help me more than anything else getting used to another keyboard-driven program with a very full keymap. (YMMV and actually I'd be interested in what helps you personally grok a new tool.)

> vim is a bit of an unfortunate example, with hjkl having special meaning there :)

Back in the day computers did not have arrow keys (and we had to walk uphill to school both ways—kidding, I'm not old enough to have witnessed this).

Other than hjkl, vi's keyboard layout mostly makes sense imo, and is a good example for keyboard-driven software (albeit taking some time getting used to), perhaps because it has had some level of influence on the rest of the *nix software ecosystem over the decades.

> For those I had in mind 'next' (so META+Escape=next screen, META+`=next ws, META+Tab=next frame, META+CapsLock=next tab)

Makes sense this way as well.

I trust you to make the right decisions, and maybe it isn't much value in "adding my own mustard" as we say here.

etc/cfg_bindings.lua
raboof: Updated to use 'X' for 'Unsplit' (similar to ion3 and makes sense since that's pretty destructive)
Updated to use 'N' ('nick') for 'attach' (a bit easier to reach than 'P')

raboof:
TODO: make sure `ALTMETA` defaults to `META+Shift` when it is set to the empty string.

etc/cfg_bindings.lua
```diff
@@ -55,13 +55,13 @@ defbindings("WScreen", {
...
+    kpress(META.."F2", "ioncore.goto_nth_screen(1)"),
     kpress(META.."Shift+3", "ioncore.goto_nth_screen(2)"),
-    kpress(META.."F2", "ioncore.goto_nth_screen(2)"),
-    kpress(META.."Shift+4", "ioncore.goto_nth_screen(2)"),
     kpress(META.."F3", "ioncore.goto_nth_screen(2)"),
+    kpress(META.."Shift+4", "ioncore.goto_nth_screen(3)"),
+    -- kpress(META.."F4", "ioncore.goto_nth_screen(3)"),
 
     bdoc("Go to next/previous screen on multihead setup."),
     kpress(ALTMETA.."Escape", "ioncore.goto_prev_screen()"),
```
wilhelmy: In WMoveResize mode, AnyModifier+Escape is bound to cancel resizing.. Does this still work if  (ALT)META.."Escape" is bound to switch screens?

raboof: it does look like it (but I don't have a second monitor handy to *really* test this right now)

raboof:
I think this is ready for final review/merge, with further evolution be done as subsequent PR's.

knixeur:
LGTM

Issue #144, opened 2019-07-02
Consider replacing `run-mailcap` by `xdg-open` in the default config


wilhelmy:
Should be more portable and use the system's standard MIME database instead of yet another one (I assume run-mailcap uses `~/.mailcap`)

knixeur:
I agree, xdg-open has been around for a long time and works fine.

wilhelmy:
The only question left to answer is whether it has an equivalent of `run-mailcap --action=edit`.

knixeur:
Mmm you're right, I seldome use this options, maybe Edit can be dropped since editable stuff is usually opened with it's editor (IE: png->gimp, c->vim) the most common problematic cases I can think of are html files.

raboof:
I see a bare Arch install doesn't install run-mailcap but does seem to have xdg-open.

Should we switch to xdg-open or have both options side by side?

wilhelmy:
Maybe xdg-open for open, run-mailcap for edit?

PR #162, opened 2019-07-31
Generate keybinding documentation

raboof:
So we can show docs like https://notionwm.net/notionkeys.html that actually correspond to the current configuration

### DIFF
--- a/ioncore/ioncore_misc.lua
+++ b/ioncore/ioncore_misc.lua
@@ -99,7 +99,7 @@ function ioncore.tagged_attach(reg, param)
 end
 
 function ioncore.show_live_docs(frame)
-    keysfile = io.open("/tmp/notionkeys.html", "w")
+    local keysfile = io.open("/tmp/notionkeys.html", "w")
     io.output(keysfile)
 
     io.write([[
@@ -248,16 +248,12 @@ $().ready(function(){
 
 ]])
 
-    bindings_by_key = {}
+    local bindings_by_key = {}
 
-    function bindings_for(label)
-        --io.write(label .. " commands:\n")
+    local function bindings_for(label)
         for _,binding in pairs(notioncore.do_getbindings()["W" .. label]) do
-            if binding["doc"] then
-                --io.write(binding["kcb"] .. ": " .. binding["doc"] .. "\n")
-                if binding["label"] then
-                    bindings_by_key[binding["kcb"]] = binding["label"]
-                end
+            if binding["doc"] and binding["label"] then
+                bindings_by_key[binding["kcb"]] = binding["label"]
             end
         end
     end
@@ -269,7 +265,7 @@ $().ready(function(){
     bindings_for("Tiling")
     bindings_for("MPlex.toplevel")
 
-    function binding_label(label)
+    local function binding_label(label)
         if label:sub(1,2) == "->" then
             return "&rarr;" .. label:sub(3,#label)
         elseif label:sub(1,2) == "<-" then
@@ -291,7 +287,7 @@ $().ready(function(){
         end
     end
 
-    function print_key(label, class, name)
+    local function print_key(label, class, name)
         if name == nil then
             name = label
         end
@@ -333,7 +329,7 @@ $().ready(function(){
     print_key("Backspace", "backspace lastitem", "BackSpace")
 
     print_key("Tab", "tab")
-    keys = "QWERTYUIOP"
+    local keys = "QWERTYUIOP"
     for i=1,#keys do
         print_key(keys:sub(i,i))
     end
@@ -368,14 +364,13 @@ $().ready(function(){
     io.write("All keybindings are activated by also pressing the META key, which by default is the 'windows' key.<br>\n")
     io.write("The 'grey' bindings are activated by pressing Shift in addition to META.<br><br>\n")
 
-
-    function list_keys_for(label, ctx)
+    local function list_keys_for(label, ctx)
         if label then
             io.write("<tr><th colspan=\"2\">" .. label .. "</th></tr>\n")
         end
         for _,binding in pairs(notioncore.do_getbindings()[ctx]) do
             if binding["doc"] then
-                ref = binding["kcb"]
+                local ref = binding["kcb"]
                 if ref:sub(1,6) == "Shift+" then
                     ref = ref:sub(7, #ref)
                 end
--- a/ioncore/ioncore_misc.lua
+++ b/ioncore/ioncore_misc.lua
@@ -100,9 +100,8 @@ end
 
 function ioncore.show_live_docs(frame)
     local keysfile = io.open("/tmp/notionkeys.html", "w")
-    io.output(keysfile)
 
-    io.write([[
+    keysfile:write([[
 <html>
 <head>
   <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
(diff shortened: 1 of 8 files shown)


ioncore/ioncore_misc.lua
```diff
@@ -98,3 +98,319 @@ function ioncore.tagged_attach(reg, param)
...
+<body>
+<h1>Notion keyboard reference:</h1>
+
+<!-- CSS keyboard from http://net.tutsplus.com/tutorials/javascript-ajax/creating-a-keyboard-with-css-and-jquery/ -->
+
+]])
+
+    bindings_by_key = {}
+
+    function bindings_for(label)
```
wilhelmy: local function?

ioncore/ioncore_misc.lua
```diff
@@ -98,3 +98,319 @@ function ioncore.tagged_attach(reg, param)
...
+</script>
+</head>
+<body>
+<h1>Notion keyboard reference:</h1>
+
+<!-- CSS keyboard from http://net.tutsplus.com/tutorials/javascript-ajax/creating-a-keyboard-with-css-and-jquery/ -->
+
+]])
+
+    bindings_by_key = {}
```
wilhelmy: local bindings_by_key

ioncore/ioncore_misc.lua
```diff
@@ -98,3 +98,319 @@ function ioncore.tagged_attach(reg, param)
...
+    end
+    bindings_for("ClientWin")
+    bindings_for("Screen")
+    bindings_for("Frame")
+    bindings_for("Frame.toplevel")
+    bindings_for("GroupCW")
+    bindings_for("Tiling")
+    bindings_for("MPlex.toplevel")
+
+    function binding_label(label)
```
wilhelmy: local

ioncore/ioncore_misc.lua
```diff
@@ -98,3 +98,319 @@ function ioncore.tagged_attach(reg, param)
...
+
+    io.write("<table class=\"keytable\" style=\"clear: left\">\n")
+    io.write("<tbody><tr>\n")
+    io.write("<td width=\"400px\" valign=\"top\">\n")
+    io.write("<h2>META key</h2>\n")
+    io.write("All keybindings are activated by also pressing the META key, which by default is the 'windows' key.<br>\n")
+    io.write("The 'grey' bindings are activated by pressing Shift in addition to META.<br><br>\n")
+
+
+    function list_keys_for(label, ctx)
```
wilhelmy: local

ioncore/ioncore_misc.lua
```diff
@@ -98,3 +98,319 @@ function ioncore.tagged_attach(reg, param)
...
+        for _,binding in pairs(notioncore.do_getbindings()[ctx]) do
+            if binding["doc"] then
+                ref = binding["kcb"]
+                if ref:sub(1,6) == "Shift+" then
+                    ref = ref:sub(7, #ref)
+                end
+                if ref:sub(1,5) == "Mod4+" then
+                    ref = ref:sub(6, #ref)
+                end
+                io.write("<tr class=\"shortcut\" ref=\"#" .. ref .. "\"><td>" .. binding["kcb"] .. "</td><td>" .. binding["doc"] .. "</td></tr>\n")
```
wilhelmy: I guess you could save some backslashes by using single quotes?

ioncore/ioncore_misc.lua
```diff
@@ -98,3 +98,319 @@ function ioncore.tagged_attach(reg, param)
...
+        elseif label=="hmax" then
+            return "&harr;max"
+        elseif label=="vmax" then
+            return "&varr;max"
+        else
+            return label
+        end
+    end
+
+    function print_key(label, class, name)
```
wilhelmy: local

ioncore/ioncore_misc.lua
```diff
@@ -98,3 +98,319 @@ function ioncore.tagged_attach(reg, param)
     return true
 end
 
+function ioncore.show_live_docs(frame)
+    keysfile = io.open("/tmp/notionkeys.html", "w")
```
wilhelmy: local

ioncore/ioncore_misc.lua
```diff
@@ -98,3 +98,319 @@ function ioncore.tagged_attach(reg, param)
...
+    io.write("</td>")
+    io.write("</tr></tbody>\n")
+    io.write("</table>\n")
+
+    io.write("</body>")
+    io.write("</html>")
+
+    io.close(keysfile)
+
+    mod_query.exec_on_merr(frame, "run-mailcap --mode=view /tmp/notionkeys.html")
```
wilhelmy: maybe xdg-open instead of run-mailcap? AFAIK the only good reason for run-mailcap is that it has a `--mode=edit`, but I think it also only exists on debian at all?

ioncore/ioncore_misc.lua
```diff
@@ -98,3 +98,319 @@ function ioncore.tagged_attach(reg, param)
...
+    print_key("/", nil, "slash")
+    io.write("<li class=\"right-shift lastitem\">shift</li>\n")
+    io.write("</ul></div>")
+
+    io.write("<table class=\"keytable\" style=\"clear: left\">\n")
+    io.write("<tbody><tr>\n")
+    io.write("<td width=\"400px\" valign=\"top\">\n")
+    io.write("<h2>META key</h2>\n")
+    io.write("All keybindings are activated by also pressing the META key, which by default is the 'windows' key.<br>\n")
+    io.write("The 'grey' bindings are activated by pressing Shift in addition to META.<br><br>\n")
```
knixeur: I think this should read "The 'grey' bindings are activated by pressing ALTMETA (by default pressing Shift while holding META.<br><br>\n"

ioncore/ioncore_misc.lua
raboof: I agree, but 'view' also uses 'run-mailcap' right now. Let's keep it consistent but move to xdg-open in #144

raboof:
Great review comments! Should all be done now.

wilhelmy:
I think there's some cases of missing 'local' keywords left.. I've seen `keys = 'foo'` at least.

ioncore/ioncore_misc.lua
```diff
@@ -98,3 +98,319 @@ function ioncore.tagged_attach(reg, param)
     return true
 end
 
+function ioncore.show_live_docs(frame)
+    keysfile = io.open("/tmp/notionkeys.html", "w")
+    io.output(keysfile)
```
wilhelmy: I think instead of io.output/io.write (I wasn't even aware of io.output... the more you know...) it would be better to use keysfile:write(), which does not modify global state?

ioncore/ioncore_misc.lua
wilhelmy: Fair enough. Not sure what we should do about the "edit" feature, which xdg-open does not have...

wilhelmy:
I guess otherwise it's good to go. Cool new feature!

wilhelmy:
> I think there's some cases of missing 'local' keywords left.. I've seen `keys = 'foo'` at least.

Disregard this, of course you only need to declare the first assignment local.

ioncore/ioncore_misc.lua
raboof: Nice one, thx

ioncore/ioncore_misc.lua
```diff
@@ -100,9 +100,8 @@ end
 
 function ioncore.show_live_docs(frame)
     local keysfile = io.open("/tmp/notionkeys.html", "w")
```
wilhelmy: This could break if more than one user on the system uses notion, or if another user created the file already. Not sure it's really a concern.

raboof: Changed this to use `os.tmpname()` (as `io.tmpfile()` does not necessarily have a name). I guess ideally we should clean up those files when exiting, but at least on linux /tmp will be cleared on reboot anyway, so this might be fine for now.

wilhelmy: Good to know, I only knew about `io.tmpfile()`. The only problem is that the filename doesn't end with .html now... So maybe it might be required to call `os.rename(filename, filename..".html")` before opening, but if you say it works, I'll believe you :)

raboof: It works with firefox, but it's a good suggestion to add the `.html`, will do that in a follow-up PR

PR #166, opened 2019-08-06
Make sure live docs have a '.html' extension

raboof:
Not needed for firefox at least, but neater

### DIFF
--- a/ioncore/ioncore_misc.lua
+++ b/ioncore/ioncore_misc.lua
@@ -99,7 +99,7 @@ function ioncore.tagged_attach(reg, param)
 end
 
 function ioncore.show_live_docs(frame)
-    local filename = os.tmpname() .. ".html"
+    local filename = ioncore.tempdir() .. ".html"
     local keysfile = io.open(filename, "w")
 
     keysfile:write([[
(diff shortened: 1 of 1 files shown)


ioncore/ioncore_misc.lua
```diff
@@ -99,7 +99,7 @@ function ioncore.tagged_attach(reg, param)
 end
 
 function ioncore.show_live_docs(frame)
-    local filename = os.tmpname()
+    local filename = os.tmpname() .. ".html"
```
wilhelmy: A problem here is that at least with Lua 5.3, os.tmpname() also creates the file instead of just returning a unique name (I guess because otherwise it can't be guaranteed that the filename is unique).

knixeur: It actually depends on the OS implementation. The alternative is to include posix Lua module but that would add a requirement just for this. I think 2 temp files won't hurt very much, it's not a functionality that will be run that often I think

wilhelmy: I propose using `local filename = ioncore.tempdir().."notionflux.html"` from my upcoming pull request.

raboof:
Thanks!

Issue #114, opened 2019-06-18
New default keybindings


raboof:
I think Notion 4 should come with a new set of keybindings.
* More consistent/memorable
* Navigate left-handed, query/rearrange(?) right-handed.
* Replace `META+K` with `ALTMETA`, and define `ALTMETA` as `META+Shift`.
* Keep it easy to switch back to the old keybindings

Probably:
* [x] Create a new `cfg_bindings.lua` with the new default bindings
* x Also put the (new) 'cfg_tiling.lua' bindings in `cfg_bindings.lua`.
* [x] Put old keybindings in `cfg_bindings_notion3.lua`
* [x] Make `cfg_notioncore.lua` empty aside from including `cfg_bindings.lua`
* [x] Put the new default bindings in `cfg_bindings.lua`
* [x] Add to the migration guide #113 that you can switch back by copying `cfg_bindings_notion3.lua` as `~/.notion/cfg_bindings.lua`
* [ ] Make sure `ALTMETA=""` is interpreted as `ALTMETA=META.."Shift+"`

wilhelmy:
Should META+Button1 behave as "undock and move the resulting floating window around" on tiled windows? Currently it's used for resize, but META+Button2 can also be used to resize both float and tiled windows.

rabarbar42:
Hi. I just used git to get the latest Notion snapshot, compiled it and ...I totally don't recognize it anymore. To begin with, how to switch back to old keybindings? Then, META+K doesn't work, numbered tabs don't work (either with meta or altmeta), and hide_tabs is broken...

...and there's no email list anymore
...and I could go on for a long time.

raboof:
Thanks for bringing it up! (I see @knixeur also replied while I was typing this, so there might be some overlap ;) )

We're gearing up for a "Notion 4", and indeed we'd like to make some bigger changes.

Of course that should not be a "just put up with it" thing for existing users: while some of the defaults change, for most things it should ideally be possible to get the "old" behavior back with some configuration. For this we started on a migration guide which can be found at https://notionwm.net/migration.html (or, if you want to contribute to it, https://github.com/raboof/notion/blob/gh-pages/migration.md

> To begin with, how to switch back to old keybindings?

[...]

PR #172, opened 2019-08-27
Document reverting to the old keybindings

raboof:
Refs #114


notion4keys.html
```diff
@@ -0,0 +1,304 @@
+<html>
```
wilhelmy: Wait, I thought the keybinding document is generated? Or is this the gh-pages branch?

wilhelmy: Nvm, it is.

PR #174, opened 2019-09-02
add flag -nodefaultconfig to prevent loading system default dirs

dradetsky:
I made this to support another contribution I have (a handy, ready-to-go xephyr-for-dev setup), although I suppose it could have other applications.

This adds a flag `-nodefaultconfig` which prevents notion from loading the standard directories into its search path.

I've tested the basic cases manually (i.e. no arguments, one/two `-s` flags with/without the new flag), but it looks like the unit test setup doesn't really work, so I haven't tried to add any tests.

### DIFF
--- a/notion/notion.c
+++ b/notion/notion.c
@@ -60,6 +60,9 @@ static OptParserOpt ion_opts[]={
     {OPT_ID('N'), "noerrorlog", 0, NULL,
      DUMMY_TR("Do not create startup error log and display it "
               "with xmessage.")},
+    {OPT_ID('Z'), "nodefaultconfig", 0, NULL,
+     DUMMY_TR("Do not add the default directories to the "
+              "search path.")},
 
     {'h', "help", 0, NULL,
      DUMMY_TR("Show this help")},
@@ -149,18 +152,12 @@ int main(int argc, char*argv[])
     char *efnam=NULL;
     bool may_continue=FALSE;
     bool noerrorlog=FALSE;
+    bool nodefaultconfig=FALSE;
 
     libtu_init(argv[0]);
 
     if(!ioncore_init("notion", argc, argv, LOCALEDIR))
         return EXIT_FAILURE;
-
-    extl_add_searchdir(EXTRABINDIR); /* ion-completefile */
-    extl_add_searchdir(MODULEDIR);
-    extl_add_searchdir(ETCDIR);
-    extl_add_searchdir(SHAREDIR);
-    extl_add_searchdir(LCDIR);
-    extl_set_userdirs("notion");
 
     optparser_init(argc, argv, OPTP_MIDLONG, ion_opts);
 
@@ -187,6 +184,9 @@ int main(int argc, char*argv[])
         case OPT_ID('N'):
             noerrorlog=TRUE;
             break;
+        case OPT_ID('Z'):
+            nodefaultconfig=TRUE;
+            break;
         case 'h':
             help();
             return EXIT_SUCCESS;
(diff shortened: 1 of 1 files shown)


dradetsky:
@raboof I'm not going to pretend I entirely understand what those directories are all for, but to the extent that I understand, no, that's not what we want. For example, in dev I want to use the compiled lua modules from the git repo, not the system.

As I recall (maybe this was changed?) last time I tried this it didn't work like this and I ended up having to comment out the relevant block of standard loads to get it to work. For example, I think it expects all the `.lc` files to be in `$prefix/notion/lc` (`/usr/lib/notion/lc` in my case), whereas the build system actually sticks them all in the various module directories. So if I want to load them from the dev repo only, I can't just reset `$prefix`.

Maybe there's a better way to do this, but it's not specified anywhere I was able to find easily.

dradetsky:
@raboof fwiw, I could add a flag that skips just `~/.notion` and a flag that skips everything on the theory that the former is something that most people want & it's related to what I'm doing now. But I didn't really think about it because as I said, I'm pretty sure skipping everything is required for my use-case. Also, I thought maybe adding tons of flags would be a bad thing. But if you don't think so, then I guess it isn't.

raboof:
> For example, in dev I want to use the compiled lua modules from the git repo, not the system

Ah, you want to test a freshly built notion without doing 'make install'. I agree this is desirable, and not supported right now.

> I could add a flag that skips just ~/.notion and a flag that skips everything

I think we should *always* load `~/.notion` - if someone doesn't want that he can simply point the `HOME` environment variable somewhere else.

Then there is 'everything else'. To support testing a freshly built notion without doing 'make install', I think we would not only need to skip loading things from `/usr`, but we should also make sure the scripts from inside your notion checkout *are* loaded, right?

dradetsky:
> I think we should always load ~/.notion - if someone doesn't want that he can simply point the HOME environment variable somewhere else.

That's not a terrible idea, IMO, but here's a downside: suppose I don't know whether that env var is used for anything else by notion? As a matter of fact, I don't, since I didn't check the code. So I don't what to set home to something else, because I have no idea what that will affect, if anything. So that solution won't naturally occur to me.

In a sense, you've also added a weird dependency on HOME: it can't be part of the interface in any other way, because you have to be able to use it to determine where you load configuration from. So if somebody in the future wants to add a feature which uses HOME for something, they can't. Or, they can, but from the perspective of notion HOME isn't actually the user's home dir, it's wherever the user wants to load the user-level notion config from.

It was confusing just writing that, so I can't imagine it will be easy to explain to all future contributors. By contrast "load absolutely nothing" "load this specific stuff" is very easy to understand.

FWIW, I've recently done some work with emacs customization where the assumption that `~/.emacs.d` is always an appropriate load path is really annoying. It's annoying _even_ with the ability to say "don't load any of the default code". It may be a bad comparison, but it would be even worse if this option wasn't available.

> To support testing a freshly built notion without doing 'make install', I think we would not only need to skip loading things from /usr, but we should also make sure the scripts from inside your notion checkout are loaded, right?

I agree in general, but not necessarily in practice. For example, if I run a version of notion that's reasonably close to the tip of master, I may not care about this stuff. It may not be relevant to use the checkout lua. I may just be changing the C. Also, doing this would presumably require changing the way we do the build, or the install, or both. Which I didn't know how to do. And which could presumably break downstream packages across every linux distro. I get that this has been a non-goal of the notion project since the Tuomo (PBUH) era, but it's still not a good thing. I'd only want to do it (i.e. change the build/install) if I knew how to do it properly.

With this change, you at least can do what you need to do for dev without having to comment out C code and rebuild. And you can fix it for developers in general in the short run by adding a utility wrapper script and saying "Run notion with `notion $(./dev-search-path-flags.sh) -other -flags`." And it doesn't _prevent_ you from adding some more sophisticated way of ensuring that the appropriate stuff is loaded from a checkout.

I suppose I could add a note to the flag docs with "NOTE: this will break notion unless you add the appropriate -s flags to other places".

wilhelmy:
Thanks, I've been running into this exact same issue (installed into a temporary DESTDIR and tried to run from there without using anything in /usr{,/local}) and if we were going to vote, my vote would be in favour of your approach instead of misusing $HOME.

raboof:
I didn't realize before that setting HOME will also affect any programs running 'under' notion, so on second thought I agree that is not a nice way to prevent loading your custom config.

> presumably break downstream packages across every linux distro. I get that this has been a non-goal of the notion project since the Tuomo (PBUH) era, but it's still not a good thing

Being 'downstream-friendly' is definitely a goal now, and a good part of the reason we've been putting in the work to make Notion properly LGPL again.

> it doesn't prevent you from adding some more sophisticated way of ensuring that the appropriate stuff is loaded from a checkout

Right, I agree this is a good first step, and users can use `-s` to add source directories as needed. Some documentation (or even scripts) on how to do that for common scenario's might be nice, but that could be a subsequent PR.

notion/notion.c
```diff
@@ -203,6 +203,27 @@ int main(int argc, char*argv[])
         }
     }
 
+    char *tmp_searchpath=NULL;
```
raboof: So this is done to make sure the search directories added with `-s` are appended to the search path rather than prepended, because the most recently added path has the highest precedence? That makes sense I guess.

knixeur:
Great work, thanks!

Issue #154, opened 2019-07-12
Make sure ALTMETA defaults to META+Shift when it is set to the empty string


raboof:
Perhaps edit the notion3 fallback bindings to not use ALTMETA (since the default was the empty string there)

#114

raboof:
Still to do: mention in the migration notes that `ALTMETA` should no longer be empty, but rather something like `Mod4+Shift+`

raboof:
documented!

PR #179, opened 2019-09-13
Default to Mod4+Shift for altmeta

raboof:
Refs #154

Producing a better error message when you have it set to ""


raboof:
Let's also add this to the migration docs before closing this

PR #192, opened 2019-09-25
Change references from sourceforge to github

knixeur:
Should we add  these to notionwm? Is it possible?


raboof:
> Should we add these to notionwm? Is it possible?

Yes, but it's quite a bit of work. I thought there was an issue for this but can't find it, so I created https://github.com/raboof/notion/issues/193

PR #194, opened 2019-09-28
(Re?)enable navigation bar and populate it from *.md

wilhelmy:


wilhelmy:
I have a fix for that in a subsequent PR which restructures the site

PR #195, opened 2019-09-28
Change site structure, add a little more content

wilhelmy:
Please expand :)

### DIFF
--- a/contact.md
+++ b/contact.md
@@ -11,7 +11,7 @@ longer very active.
 We use the [github issue tracker](https://github.com/raboof/notion/issues) and
 pull requests for code contribution.
 
-For discussion about the future of the project, we either use issues tagged "enhancement".
+For discussion about the future of the project, we use issues tagged "enhancement".
 
 For support questions and notion related chat such as quickly asking a question,
 we use IRC:
--- a/contact.md
+++ b/contact.md
@@ -1,0 +1,20 @@
+---
+layout: default
+title: Contact
+---
+# Contact
+
+There is a [mailing list](https://sourceforge.net/p/notion/mailman/notion-devel/)
+on sourceforge, but ever since the project has migrated to github, it is no
+longer very active.
+
+We use the [github issue tracker](https://github.com/raboof/notion/issues) and
+pull requests for code contribution.
+
+For discussion about the future of the project, we either use issues tagged "enhancement".
+
+For support questions and notion related chat such as quickly asking a question,
+we use IRC:
+#notion on Freenode ([Webchat](https://webchat.freenode.net/?channels=%23notion)).
+
+Please don't hesitate to get in touch, usually IRC is the easiest way.
--- a/contact.md
+++ b/contact.md
@@ -4,10 +4,6 @@ title: Contact
 ---
 # Contact
 
-There is a [mailing list](https://sourceforge.net/p/notion/mailman/notion-devel/)
-on sourceforge, but ever since the project has migrated to github, it is no
-longer very active.
-
 We use the [github issue tracker](https://github.com/raboof/notion/issues) and
 pull requests for code contribution.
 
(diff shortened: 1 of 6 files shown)


contact.md
```diff
@@ -0,0 +1,20 @@
+---
+layout: default
+title: Contact
+---
+# Contact
+
+There is a [mailing list](https://sourceforge.net/p/notion/mailman/notion-devel/)
+on sourceforge, but ever since the project has migrated to github, it is no
+longer very active.
```
raboof: Perhaps we shouldn't even mention it, or just at the bottom

contact.md
```diff
@@ -0,0 +1,20 @@
...
+
+We use the [github issue tracker](https://github.com/raboof/notion/issues) and
+pull requests for code contribution.
+
+For discussion about the future of the project, we either use issues tagged "enhancement".
```
raboof:
```suggestion
For discussion about the future of the project, we use issues tagged "enhancement".
```

contact.md
```diff
@@ -0,0 +1,20 @@
...
+
+For support questions and notion related chat such as quickly asking a question,
```
raboof: Perhaps we should clarify that when there's no quick reply on IRC, opening an issue would be a good next step?

index.md
```diff
@@ -35,9 +36,7 @@ For writing patches and modules, see [Notes for the module and patch writer](htt
 
 # Support, bugs and feature requests
 
-Find us on IRC, in #notion on freenode.
-
-Report your bugs and feature requests [in the GitHub trackers](https://github.com/raboof/notion/issues).
+See [Contact](contact.html).
```
raboof: :+1: - though perhaps we should remove the whole section here?

software.md
```diff
@@ -0,0 +1,56 @@
...
+
+These trays can be used to add tray icon functionality either to Notion's integrated status bar (although we recommend using a panel, most of which include support for tray icons, see below).
+
+* [stalonetray](http://stalonetray.sourceforge.net/) shows FreeDesktop (XEmbed) and KDE trayicons
+* [trayion](https://code.google.com/archive/p/trayion/) shows FreeDesktop (XEmbed) trayicons (like Qt4 and gtk2), integrates well with <s>Not</s>ion's built-in status bar (mod\_statusbar), but seems abandoned. Can embed itself into mod\_statusbar.
+
+## Panels
+
+* [xfce4-panel](https://docs.xfce.org/xfce/xfce4-panel/start) — the classic panel from XFCE. Includes various widgets, most notably volume control, battery, clock, disk usage, system tray, application launcher.
+* [polybar](https://polybar.github.io/) — A status bar written in C++, supporting modules and .ini-file customization. Plugins working with notion include volume control, clock, filesystem usage, keyboard layout, memory usage, cpu usage, battery, temperature
```
raboof: :+1:

contact.md
wilhelmy: I mostly decided to mention it in case someone wants to look at the archive, but I guess I could have explained that.

index.md
wilhelmy: The reason I left it in was because the heading sounds more friendly than simply "Contact", which people might mistake for "Imprint (for legal reasons)".

contact.md
wilhelmy: I've had it happen that people who aren't familiar with IRC would join and quickly quit again, or join, ask a question and wait for a while, but then leave... so maybe you're right. At the same time handling the entire communication through github issues seems a bit weird to me, maybe that's another reason why I left the link to the mailing list in, because my head associates issue tracker with issues that need fixing, and not general project communication. Maybe github is trying to evolve the issue tracker into a generic communication tool, and of course we're free to use it, but I think we should also explain that in this case.

In a way, the Contact document was intended as an intro to how we communicate in this project. If you have suggestions, please just edit it.

index.md
wilhelmy: However, due to alphabetical sorting, "Contact" comes before "Home" in the navigation bar, therefore emphasizing our willingness to communicate with the user base enough that we can probably delete the passage ;)

PR #196, opened 2019-10-01
Code of conduct

raboof:
I wonder if we should have a code of conduct.

While the Notion community seems a bit too small to have a 'full-fledged' code of conduct like some larger projects have, it might be nice to put at least something in writing.

What do you think about something like this?

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -16,7 +16,7 @@ So far we have only scratched the surface of this potential.
 ## On governance
 
 Notion is a relatively small community. At this scale it doesn't seem worth it
-yet to set up a governence structure as found in larger communities, so until
+yet to set up a governance structure as found in larger communities, so until
 then, raboof is acting as the BDFN: "Benevolent Dictator For Now".
 We will revisit this choice when the community grows.
 
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -58,4 +58,4 @@ way to resolve the situation.
 
 We recognize that misunderstandings happen and well-meant actions may have
 unintended effects. We will only take actions when a user appears unwilling or
-unable to sufficiently adjust his behaviour after repeated conversations.
+unable to sufficiently adjust their behaviour after repeated conversations.
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -16,24 +16,46 @@ So far we have only scratched the surface of this potential.
 ## On governance
 
 Notion is a relatively small community. At this scale it doesn't seem worth it
-yet to set up a governance structure as found in larger communities, so until
-then, raboof is acting as the BDFN: "Benevolent Dictator For Now".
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,39 @@
...
+respectful to all involved. If anyone feels unwelcome, disrespected or even
+threatened by someone else's behaviour, this means we have something to
+improve. The emphasis should not be on punishing bad behaviour, but on finding
+ways to take better consideration of each other going forward.
+
+If you feel uncomfortable voicing concerns around conduct publically, feel free
+to contact raboof. He promises to treat such issues with respect and privacy,
+and work with you and any others to find a way to resolve the situation.
+If this contact point does not work for you we unfortunately don't have any
+alternatives available at this time.
```
wilhelmy: Hmmm... "Pick someone from the community who you trust, and raise the issue with them" might be better than nothing?

wilhelmy:
While I don't think a CoC is necessary, I have no objections against either of the two, leaning towards raboof's because that means it will be easier to make changes later on because it feels more like we're personally in charge as a community.

However a CoC is useless if not enforced. I'm not sure blindly slapping one on every project accomplishes anything more than showing good intentions.

raboof:
> "Pick someone from the community who you trust, and raise the issue with them" might be better than nothing?
> However a CoC is useless if not enforced. I'm not sure blindly slapping one on every project (which seems to be a trend) accomplishes anything more than showing good intentions.

I agree completely: if we add a CoC, we should take it seriously and make sure we can act on it when necessary. This is why I didn't add "Pick someone from the community who you trust, and raise the issue with them": I wanted to be clear that a "shortcoming" of our approach is that if I'm misbehaving, there currently is no way to enforce the CoC against me. Introducing the proper structure for this would not really make sense for a community as small as ours, so that's a limitation we accept for now. I'll see if I can find a way to word that more clearly without making too much of a 'thing' of it.

ideasman42:
One of the things I find difficult about this CoC is, after reading, I don't have a clear idea of what is/isn't OK.
there is brief statement about needing to be respectful, but it's mixed in with general statements which aren't actionable.

A CoC is fine but it's putpose should be clear, I should be able to be aware of it and move on with life :) anything thats un-clear or vague, gets in the way and is annoying.

For this reason, suggest following Debian/Gnome/Blender's CoC because it's short, no-nonsense guidelines are easy to remember and act on, they're also fairly common so people don't have to remember too many different CoC's when working on different projects.

raboof:
> For this reason, suggest following Debian/Gnome/Blender's CoC because it's short, no-nonsense guidelines are easy to remember and act on, they're also fairly common so people don't have to remember too many different CoC's when working on different projects.

Those guidelines seem reasonable, I have updated the text with them.

ideasman42:
In general LGTM, although I'd personally remove all non-essential text _(when it comes to rules "less is more" IMHO)_, or move elsewhere (some of it reads more like a mission statement, not actually guiding peoples behavior).

PR #198, opened 2019-10-05
Cleanup: quiet '-Wextra-semi-stmt' clang warning

ideasman42:
Ensures macros end with a semi-colon,
avoids accidents with macro use.

### DIFF
--- a/ioncore/event.c
+++ b/ioncore/event.c
@@ -66,7 +66,7 @@ static void check_signals()
 
 /*{{{ Timestamp stuff */
 
-#define CHKEV(E, T) case E: tm=((T*)ev)->time; break;
+#define CHKEV(E, T) case E: tm=((T*)ev)->time; break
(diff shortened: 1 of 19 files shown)


ioncore/event.c
```diff
@@ -66,7 +66,7 @@ static void check_signals()
 
 /*{{{ Timestamp stuff */
 
-#define CHKEV(E, T) case E: tm=((T*)ev)->time; break;
+#define CHKEV(E, T) case E: tm=((T*)ev)->time; break; ((void)0)
```
raboof: In cases where the macro is a statement rather than a block, wouldn't it be neater to just drop the final `;` rather than appending a `((void)0)`?
```suggestion
#define CHKEV(E, T) case E: tm=((T*)ev)->time; break
```

ideasman42: Done.

PR #207, opened 2019-10-08
Encourage getting Notion from git

raboof:


software.md
```diff
@@ -2,6 +2,10 @@
 layout: default
 title: Software
 ---
+# Download Notion
+
+Get _**Not**ion_ from our [git repo](https://github.com/raboof/notion)
```
wilhelmy: Should we say something about the stability of master?

raboof: I think people who can clone git understand the risk, and our master is generally pretty stable. WDYT?

wilhelmy: That's my point, since master is pretty stable we can encourage people to try it without feeling bad about it :)

knixeur: I think we're all in the same page :D
I was reading software.md, there we should encourage to use panels (move it first)  instead of mentioning systray apps since we're deprecating statusbar.

wilhelmy: I think we can safely use the gh-pages branch like a wiki, as in edit and fire, instead of submitting PRs :)

raboof: That's fine by me

PR #209, opened 2019-10-10
Change query_runfile (View) to use xdg-open by default

### LINKED ISSUE #144: Consider replacing `run-mailcap` by `xdg-open` in the default config

knixeur:
Both Edit/View file can be customized by setting EDIT_COMMAND/VIEW_COMMAND
in etc/cfg_notion.lua

Fix #144

### DIFF
--- a/etc/cfg_notion.lua
+++ b/etc/cfg_notion.lua
@@ -19,6 +19,10 @@
 
 -- Terminal emulator.
 --XTERM="xterm"
+-- Edit file command, mod_query.query_editfile
+--EDIT_COMMAND="run-mailcap --action=edit"
+-- View file command, mod_query.query_viewfile
+--VIEW_COMMAND="xdg-open"
 
 -- Some basic settings.
 ioncore.set{
--- a/mod_query/mod_query.lua
+++ b/mod_query/mod_query.lua
@@ -617,12 +617,12 @@ end
 
 --DOC
 -- Asks for a file to be viewed. This script uses
--- \command{run-mailcap --action=view} by default, but you may provide an
+-- \command{xdg-open} by default, but you may provide an
 -- alternative script to use. The default prompt is "View file:" (translated).
 function mod_query.query_runfile(mplex, script, prompt)
     mod_query.query_execfile(mplex,
                              prompt or TR("View file:"),
-                             script or "run-mailcap --action=view")
+                             script or "xdg-open")
 
 end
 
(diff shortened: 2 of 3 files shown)


etc/cfg_notion.lua
```diff
@@ -19,6 +19,10 @@
 
 -- Terminal emulator.
 --XTERM="xterm"
+-- Edit file command, mod_query.query_editfile
+--EDIT_COMMAND="run-mailcap --action=edit"
+-- View file command, mod_query.query_viewfile
+--VIEW_COMMAND="xdg-open"
```
wilhelmy: JFYI, without specifying these as local, they are both global in the Lua interpreter, and available everywhere.

mod_query/mod_query.lua
```diff
@@ -617,12 +617,12 @@ end
...
 --DOC
 -- Asks for a file to be viewed. This script uses
--- \command{run-mailcap --action=view} by default, but you may provide an
+-- \command{xdg-open} by default, but you may provide an
 -- alternative script to use. The default prompt is "View file:" (translated).
 function mod_query.query_runfile(mplex, script, prompt)
     mod_query.query_execfile(mplex,
                              prompt or TR("View file:"),
-                             script or "run-mailcap --action=view")
+                             script or "xdg-open")
```
wilhelmy: … and if `VIEW_COMMAND` is global, it can theoretically be used here. Not sure that's clean design, though, and whether or not it should be global. Any opinions?

etc/cfg_notion.lua
knixeur: I know, I followed XTERM pattern, I don't know if we can/should change it.

wilhelmy: The reason I'm mentioning this is that since I started using tab completion in notionflux, I noticed a lot of identifiers polluting the global namespace, so I think we should give the "globals problem" some thought.

mod_query/mod_query.lua
knixeur: I thought about it but I saw mod_dock as a separate module and didn't want to "trust" that those variables are defined. Let me know if you think these should point to those vars.

etc/cfg_notion.lua
knixeur: We could do a "fake namespace" by using a table to hold the variables. It would bejust one global to avoid heh. Or move them to ioncore.set?

wilhelmy: Maybe we could have a "cmds" global with cmds.{XTERM,EDIT,VIEW} as children?

mod_query/mod_query.lua
wilhelmy: If we decide to go with the "cmds" approach mentioned above, I think `(cmds and cmds.VIEW) or "xdg-open"` should work around the value being nil

PR #210, opened 2019-10-10
When using XFT try loading the font with Xlfd before falling back to BMF

knixeur:


raboof:
Right - so when master renders an XLFD font, it does so with `XmbDrawImageString`/`Xutf8DrawImageString`. When this branch renders it, it does so with `XftDrawStringUtf8`/`XftDrawString8`.

While it's drawing the same font, I suspect `XftDrawStringUtf8`/`XftDrawString8` will possibly behave differently, for example applying anti-aliasing where `XmbDrawImageString`/`Xutf8DrawImageString` would not.

knixeur:
I didn't follow all the rendering logic, I was looking at something else and realized that if both BMF and XFT are enabled, XftFontOpenXlfd is never called so you always end up with BMP fonts.

raboof:
> realized that if both BMF and XFT are enabled, XftFontOpenXlfd is never called so you always end up with BMF fonts

This is a correct observation (when the font is specified in XLFD form instead of using the `xft:` prefix).

This approach is least likely to surprise people who have been using Notion from before the xft days. That seems like a good thing, no?

knixeur:
Agree :+1:

raboof:
(closing then)

PR #217, opened 2019-11-05
Don't override fonts in etc/lookcommon_*

### LINKED ISSUE #200: Configure High-DPI font for non-query menu's
what defstyle is used for those?

raboof:
So that they inherit the fonts from the including styles

This makes them more reusable across low- and high-dpi styles,
and causes less surprises with overridden fonts

fixes #200


knixeur:
If you tested it I'm pretty sure this change is positive :+1:

PR #219, opened 2019-11-06
notionflux: complete past colon

wilhelmy:
```
lua> w=ioncore.current()
nil
lua> w:<Tab>
<list of methods supported by "w" object>
```
Previously, the part before the colon would be eaten on completion attempt.

### DIFF
--- a/mod_notionflux/notionflux/notionflux.c
+++ b/mod_notionflux/notionflux/notionflux.c
@@ -287,6 +287,8 @@ char *completion_generator(const char *text, int state)
 
 	char *ret;
 	char *lastdot = strrchr(text, '.');
+	char *lastcolon = strrchr(text, ':');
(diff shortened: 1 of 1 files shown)


wilhelmy:
I think this should be good to go. If you notice other quirks with tab completion, please let me know.

mod_notionflux/notionflux/notionflux.c
```diff
@@ -287,6 +287,8 @@ char *completion_generator(const char *text, int state)
 
 	char *ret;
 	char *lastdot = strrchr(text, '.');
+	char *lastcolon = strrchr(text, ':');
+	lastdot = lastcolon && lastdot < lastcolon ? lastcolon : lastdot;
```
raboof: perhaps questionable naming but :+1: ;)

PR #243, opened 2020-01-20
Debugging

wilhelmy:
This might be useful and seems to work.


wilhelmy:
fwiw, I've tested this so it should be good to go.

raboof:
Sorry, I was out skiing. I'll give this a try and then merge soon ;)

PR #248, opened 2020-01-30
 Fix window_dialog_float not taken into account

### LINKED ISSUE #61: New "display dialog as float" feature : problem with scratchpad and option to deactivate
I have two problems with the new "display dialogs as float" feature:

- the option  "window_dialog_float=false," in etc/cfg_notion.lua does not seem to deactivate dialogs as float.
- when a dialog is opened through a window in the scratchpad, the dialog window appears *under* the scratchpad.

vincentvidal:
When window_dialog_float is false, which is the case by default, floating dialog windows are enabled anyway. REGION_PLEASE_FLOAT should not be added to region flags in that case.

fixes #61

### DIFF
--- a/ioncore/clientwin.c
+++ b/ioncore/clientwin.c
@@ -361,7 +361,7 @@ static bool clientwin_init(WClientWin *cwin, WWindow *par, Window win,
 
     netwm_update_allowed_actions(cwin);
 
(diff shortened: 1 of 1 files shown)


knixeur:
If I understand correctly, now we should not get floating dialogs by default, which we should've been getting since this feature was implemented. Just a warning, I was never bothered by it.

Issue #271, opened 2020-04-03
Version not updating


buhl:
Building from second latest commit 
```
$ notion --version
3-2019050101-653-gc8805693+20200330-0130
$
```
Looking into this, either [this line](https://github.com/raboof/notion/blob/master/version.sh#L27) needs a `--tags` argument on the git describe command
or [this line](https://github.com/raboof/notion/blob/master/Makefile#L52) needs an `--annotate` argument on the git tag command.

I can make a pull request,  but I might have overlooked something so I would like to hear what solution the project prefers, since it might affect the use of git-tags in the project.

raboof:
Thanks for noticing! I don't have any preference, what would be your recommendation?

buhl:
So after asking me for a recommendation I thought I better look into what the internet thought :)
I learned that annotated or signed tags are preferred because they come with come with information about who made the tag. I also learned that in that sense there is no difference between annotated vs signed tags which also means what is in place right now,  is working. The "problem" is that the `4.0.0` tag is not signed.

```
$  git show-ref --tags -d
47261a872f8638b12651f31da6c738b8639d97b1 refs/tags/3-2011102900
91414f3a55bad7929562d3c3486c39e82b6418d2 refs/tags/3-2012042300
19c0bba27e1b5872d7b1bb2cca9384ecf409ea41 refs/tags/3-2013030200
19c0bba27e1b5872d7b1bb2cca9384ecf409ea41 refs/tags/3-2013030201
19c0bba27e1b5872d7b1bb2cca9384ecf409ea41 refs/tags/3-2013030202
6a098aa70341b588ad4b870f590def1338adef60 refs/tags/3-2014010400
6a098aa70341b588ad4b870f590def1338adef60 refs/tags/3-2014010401
6a098aa70341b588ad4b870f590def1338adef60 refs/tags/3-2014010402
7ef817935759b1435644b66af735a2444c2d0e93 refs/tags/3-2014010500
8440b6029e87656efc453b134d2225b3d0c7b871 refs/tags/3-2014010501
f840ba9cf44e4fb9606fda7f33a3a9a3f33312de refs/tags/3-2014010502
4f86207131e4f9c695a7ce28efddd1c43cbb9ed2 refs/tags/3-2014010503
17b6418698d711dc2106b4c260296120176ad01c refs/tags/3-2014010504
1777670636da449e1ce6d71693da9622c38be5d4 refs/tags/3-2014010505
[...]

PR #273, opened 2020-04-04
Adapted nextversion.sh to support semver

buhl:
As discussed in #271 I adapted nextversion.sh and the Makefile to work with semver and `git describe`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,6 @@
 ##
 ## Notion Makefile
 ##
-REL?=p 
 
 export NOTION_RELEASE = $(shell ./version.sh)
 
@@ -48,11 +47,14 @@ snapshot:
 	cd $$DIR ;\
 	git checkout version.h
 
-dist:
-	RELEASE=`./nextversion.sh -$(REL)` ;\
-	git tag -s -m "Release $$RELEASE" $$RELEASE ; git push --tags
+RELEASE_TARGETS := $(addsuffix -dist,major minor patch)
+$(RELEASE_TARGETS):
+	@./nextversion.sh $(@) >/dev/null 2>/dev/null
+	$(eval RELEASE_VERSION=$(shell ./nextversion.sh $(@)))
+	git tag -s -m "Release $(RELEASE_VERSION)" $(RELEASE_VERSION)
+	git push --tags
 
-.PHONY: test
+.PHONY: test $(RELEASE_TARGETS)
 
 test:
 	$(MAKE) -C mod_xrandr test
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 ##
 ## Notion Makefile
 ##
+REL?=p 
 
 export NOTION_RELEASE = $(shell ./version.sh)
 
@@ -48,7 +49,7 @@ snapshot:
 	git checkout version.h
 
 dist:
-	RELEASE=`./nextversion.sh` ;\
+	RELEASE=`./nextversion.sh -$(REL)` ;\
 	git tag -s -m "Release $$RELEASE" $$RELEASE ; git push --tags
 
 .PHONY: test
--- a/Makefile
+++ b/Makefile
@@ -36,17 +36,6 @@ _install:
 		$(INSTALL) -m $(DATA_MODE) $$i $(DESTDIR)$(DOCDIR); \
 	done
 
-snapshot:
-	PWD=`pwd` ;\
-	DIR=`basename "$$PWD"` ;\
-	RELEASE=`./nextversion.sh`-snapshot ;\
-	perl -p -i -e "s/^#define NOTION_RELEASE.*/#define NOTION_RELEASE \"$$RELEASE\"/" version.h ;\
-	cd .. ;\
-	tar --exclude-vcs -czf notion-$$RELEASE-src.tar.gz $$DIR ;\
-	tar --exclude-vcs -cjf notion-$$RELEASE-src.tar.bz2 $$DIR ;\
-	cd $$DIR ;\
-	git checkout version.h
-
 RELEASE_TARGETS := $(addsuffix -release,major minor patch)
 $(RELEASE_TARGETS):
 	@./nextversion.sh $(@) >/dev/null 2>/dev/null
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ $(RELEASE_TARGETS):
 	@./nextversion.sh $(@) >/dev/null 2>/dev/null
 	$(eval RELEASE_VERSION=$(shell ./nextversion.sh $(@)))
 	git tag -s -m "Release $(RELEASE_VERSION)" $(RELEASE_VERSION)
-	git push --tags
+	@echo 'Use "git push --tags" to push the newly tagged release.'
 
 .PHONY: test $(RELEASE_TARGETS)
 
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ snapshot:
 	cd $$DIR ;\
 	git checkout version.h
 
-RELEASE_TARGETS := $(addsuffix -dist,major minor patch)
+RELEASE_TARGETS := $(addsuffix -release,major minor patch)
 $(RELEASE_TARGETS):
 	@./nextversion.sh $(@) >/dev/null 2>/dev/null
 	$(eval RELEASE_VERSION=$(shell ./nextversion.sh $(@)))
--- a/nextversion.sh
+++ b/nextversion.sh
@@ -1,18 +1,51 @@
-#!/bin/sh
+#!/bin/bash
 
-PREFIX=3-`date +"%Y%m%d"`
-
-alreadyTagged() {
-  #nfound = git tag -l | grep $1 | wc
-  return 0;
+function usage {
+    echo "Usage: $0 [-M|-m|-p]"
+    echo "    -M Major"
+    echo "    -m Minor"
+    echo "    -p patch [default]"
 }
 
-for minor in 00 01 02 03 04 05 06 07 08 09 10 11 12 13 14 15 16 17 18 19 20
-do
-  VERSION=$PREFIX$minor
-  if [ "0" = "`git tag -l | grep $VERSION | wc -l`" ]
-  then
-    echo $PREFIX$minor;
-    exit 1
-  fi
+while getopts ":Mmp" opt; do
+    case ${opt} in
+        M )
+            major=true
+            ;;
+        m )
+            minor=true
+            ;;
+        p )
+            patch=true
+            ;;
+        * )
+            usage
+            exit 1
+            ;;
+    esac
 done
+
+CURRENT_VERSION=$(git describe | cut -d- -f1)
+
+if [[ ! $CURRENT_VERSION =~ ^[1-9][0-9]?\.[0-9]+\.[0-9]+$ ]]; then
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -48,7 +49,7 @@ snapshot:
 	git checkout version.h
 
 dist:
-	RELEASE=`./nextversion.sh` ;\
+	RELEASE=`./nextversion.sh -$(REL)` ;\
 	git tag -s -m "Release $$RELEASE" $$RELEASE ; git push --tags
```
wilhelmy, line 53: I think the minus here should go into `REL?= -p`, but also a variable named `REL` in notion's main Makefile is not exactly telling with regards to its meaning.

nextversion.sh
```diff
@@ -1,18 +1,51 @@
-#!/bin/sh
+#!/bin/bash
```
wilhelmy, line 1: Please use `#!/usr/bin/env bash` so it works on systems where bash is elsewhere in $PATH (*BSD).

Makefile
buhl: I left the - out of REL because it might mess up the command lines with `make dist REL=-p`.
Could change ti to RELEASE_TYPE?
or with could make `[major|minor|patch]-dist` targets in the Makefile and skip the argument all together?

nextversion.sh
buhl: I'll change it to use env. Is env always in /usr/bin?

buhl:
How does this work? Do I just update my branch with the suggestions and the pull request will update automatically or do we need to close this one and I'll make a new one?

nextversion.sh
wilhelmy: Yes, it is.

Makefile
wilhelmy: Hm, or just use the long form `make dist REL=patch`... I'm undecided now but this looked weird to me. Maybe someone else has a better idea, I'm mostly just bikeshedding...

wilhelmy:
Yep, if you push to the branch it'll show up in here. Someone who were to review your changes can then decide whether they want to squash all commits into one commit and put that on master, do a merge commit against master or rebase all commits onto master to avoid having a merge commit cluttering the git history.

buhl:
Is the snapshot target in the Makefile still in use? If so I will adapt it to work with the newversion.sh as well. I am not sure it's working right now...

Makefile
```diff
@@ -48,11 +47,14 @@ snapshot:
 	cd $$DIR ;\
 	git checkout version.h
 
-dist:
-	RELEASE=`./nextversion.sh -$(REL)` ;\
-	git tag -s -m "Release $$RELEASE" $$RELEASE ; git push --tags
+RELEASE_TARGETS := $(addsuffix -dist,major minor patch)
+$(RELEASE_TARGETS):
```
wilhelmy, line 51: Nice one!

Makefile
```diff
@@ -47,11 +47,14 @@ snapshot:
...
-	RELEASE=`./nextversion.sh` ;\
-	git tag -s -m "Release $$RELEASE" $$RELEASE ; git push --tags
-
-.PHONY: test
+RELEASE_TARGETS := $(addsuffix -dist,major minor patch)
+$(RELEASE_TARGETS):
+	@./nextversion.sh $(@) >/dev/null 2>/dev/null
+	$(eval RELEASE_VERSION=$(shell ./nextversion.sh $(@)))
+	git tag -s -m "Release $(RELEASE_VERSION)" $(RELEASE_VERSION)
+	git push --tags
```
wilhelmy, line 55: One more thing (good thing I have origin set to my own fork). While trying to test this, I've accidentally made a notion major release: https://github.com/wilhelmy/notion/commits/5.0.0 ;)
I'd remove it since I would expect a Makefile target to tag a release but not to push it, so I could fix it locally in case of a mistake.

wilhelmy: Maybe this?
```suggestion
	@echo 'Use "git push --tags" to push the newly tagged release.
```

Makefile
```diff
@@ -47,11 +47,14 @@ snapshot:
 	cd $$DIR ;\
 	git checkout version.h
 
-dist:
-	RELEASE=`./nextversion.sh` ;\
-	git tag -s -m "Release $$RELEASE" $$RELEASE ; git push --tags
-
-.PHONY: test
+RELEASE_TARGETS := $(addsuffix -dist,major minor patch)
```
wilhelmy, line 50: I'd propose
```suggestion
RELEASE_TARGETS := $(addsuffix -release,major minor patch)
```

The way I know "make dist" from other projects, it creates a distribution tarball like "make snapshot" does with notion.

wilhelmy:
Regarding "make snapshot", I used to have relatively strong opinions that releases should come in pre-packaged tarballs signed by the maintainer, but I think we've switched to a plain tag-based distribution method with 4.0... I'm not sure if you can verify the signature on those tags without cloning the repository, though. I assume tarballs downloaded from GitHub do at least always have the same checksum.

Makefile
buhl: thanks!

Makefile
buhl: Whoops :)
I am fine with this suggestion. But the old dist target did push automagically, so this would be a change in feature.

buhl:
> Regarding "make snapshot", I used to have relatively strong opinions that releases should come in pre-packaged tarballs signed by the maintainer, but I think we've switched to a plain tag-based distribution method with 4.0... I'm not sure if you can verify the signature on those tags without cloning the repository, though. I assume tarballs downloaded from GitHub do at least always have the same checksum.

Hmm, a (very) quick tour around the internet only came up with descriptions of manual signing of tarbals.
[This one](https://wiki.debian.org/Creating%20signed%20GitHub%20releases) was referred to quite often. I don't have any experience with CI/CD tools like travis, but maybe tools like that can help?

buhl:
There might be some info related to signed releases through travis here: https://github.com/Shingyx/travis-signed-apk-release-example

wilhelmy:
I'm not a travis expert but I'll gladly steal anything ingenious that I can take away :)

wilhelmy:
Apparently it's possible to attach binary files to releases drafted using release-drafter. I think it would be possible to attach tarballs and gpg signatures there.

buhl:
> Apparently it's possible to attach binary files to releases drafted using release-drafter. I think it would be possible to attach tarballs and gpg signatures there.

Does this mean all we might have to do is tag commit and travis will then run off and make the release and sign the files? If that's the case I can remove the snapshot target from the Makefile

wilhelmy:
No, I was talking manual upload because I live in the 90s

Makefile
raboof: I'm fine with that! :)

raboof:
> I used to have relatively strong opinions that releases should come in pre-packaged tarballs signed by the maintainer, but I think we've switched to a plain tag-based distribution method with 4.0

I'm not sure signed tarballs still have value: for releases, signing the tag should be equivalent security-wise. Having a tarball separate from the git repo itself is just duplication that has the potential to be out of sync and actually hide security problems... so I would be OK with removing the `snapshot` target

wilhelmy:
The problem is, the way distribution packages work on pretty much all distributions, the build tools expect a tarball, which won't be signed even if the release is signed. Probably all hope is lost anyway, nevermind me, I'm probably gonna go `curl|sudo bash` some things later this year.

raboof:
> the way distribution packages work on pretty much all distributions, the build tools expect a tarball

Debian and Slack are happy to get the github-provided tarball (https://manpages.debian.org/stretch/devscripts/uscan.1.en.html#github.com, https://slackbuilds.org/repository/14.2/desktop/notion/), NixOS and Arch get the git tag itself (https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/window-managers/notion/default.nix#L12, https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/notion#n50), so that doesn't seem to be a big limitation at this point.

> I'm probably gonna go `curl|sudo bash` some things later this year.

I agree that is not great, but TBQH downloading a tarball, checking the signature (against what?) and running 'sudo make' is not a *lot* better really... Going offtopic here, but I would love it to be easier to compartimentalize your system more. With namespaces I think a lot of the groundwork that would make that possible is already there, but I feel 'docker' isn't really the answer yet from a usability point of view.

The convenience of NixOS's 'nix-shell' combined with the isolation provided by linux namespaces wrapped into something resembling QubesOS seems really compelling to me ;)

buhl:
So, I think this PR is as ready for a review. it says the checks are not completed but if you go to travis they are done. I think we should merge this issue and take the travis talk to a new issue, what do you guys thing?

raboof:
> take the travis talk to a new issue

I think that just 'sometimes happens', not sure if we can do anything about it.

Makefile
```diff
@@ -36,22 +36,14 @@ _install:
...
-	tar --exclude-vcs -cjf notion-$$RELEASE-src.tar.bz2 $$DIR ;\
-	cd $$DIR ;\
-	git checkout version.h
-
-dist:
-	RELEASE=`./nextversion.sh` ;\
-	git tag -s -m "Release $$RELEASE" $$RELEASE ; git push --tags
-
-.PHONY: test
+RELEASE_TARGETS := $(addsuffix -release,major minor patch)
```
raboof, line 39: I wonder if this is GNU Make-specific?

nextversion.sh
```diff
@@ -1,18 +1,44 @@
...
-  VERSION=$PREFIX$minor
-  if [ "0" = "`git tag -l | grep $VERSION | wc -l`" ]
-  then
-    echo $PREFIX$minor;
+
+CURRENT_VERSION=$(git describe | cut -d- -f1)
+
+if [[ ! $CURRENT_VERSION =~ ^[1-9][0-9]?\.[0-9]+\.[0-9]+$ ]]; then
+  (
+    echo "Error: '${CURRENT_VERSION}' is not a valid semver format";
```
raboof, line 13: I'm not sure if we want to talk about 'semver', while we use 'major.minor.patch' versioning we're not a library and don't claim to follow semver to the letter.

raboof:
Thanks, I learned a few things :D

Makefile
buhl: I am almost sure it is. Is that a problem?

nextversion.sh
buhl: How about "...is not a valid notion version format" then?

Makefile
raboof: I'm not sure if there are still platforms that Notion runs on that don't have GNU make easily available. I think it's OK to leave it like this, it can always be replaced if someone cares.

nextversion.sh
raboof: Or just 'is not a valid version format'? Anyway, no big deal - cool if you PR to fix is, but OK to leave it be as well.

Makefile
wilhelmy: We have other gmake-isms already (I remember the lua-detect.mk that I rewrote a while ago, which I'm pretty sure already contained gmake specific directives before I got started). Writing portable Makefiles is a hassle and POSIX make is very limited, it can't even include (BSD and GNU make have different syntax for include, and automake was an attempt to solve that problem by generating the Makefile instead).
I talked to NetBSD people who were of the opinion that having gmake as a build dependency for 3rd party packages is good practice.

PR #281, opened 2020-05-17
Add default implementation for WRegion:current

raboof:
This will automatically implement it for WRootWin, where the active subwindow
will typically be a WScreen.

Follow-up on #275

### DIFF
--- a/ioncore/region.c
+++ b/ioncore/region.c
@@ -248,6 +248,10 @@ WRegion *region_current(WRegion *mgr)
     return ret;
 }
 
+WRegion *region_current_default(WRegion *reg)
+{
+    return reg->active_sub;
+}
 
 void region_child_removed(WRegion *reg, WRegion *sub)
 {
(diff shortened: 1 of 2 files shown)


buhl:
I just tested this branch and the change doesn't seem to work
In notionflux
```
lua> ioncore.current():screen_of():parent():current()
nil
lua> ioncore.current():screen_of():parent():current_scr()
WScreen: 0x5609f8e0e358
lua>
```
I tried putting back this line
https://github.com/raboof/notion/pull/275/commits/a2974248bb8bebd27d93407dd01ceb12cc9b4886#diff-fc859448b8d4c58d47e4a68b962a268eR498

and it worked again.
so there is something that `region_current_default` doesn't do right when on `WRootWin`

buhl:
Now on the right branch I did this in notionflux
```
lua> repl.inspect_object(ioncore.current():screen_of():parent())
+ WRootWin: 0x55624d56c128
 + WScreen: 0x55624d56c0d8
  + WGroupWS: 0x55624d60e0e8
   + WTiling: 0x55624d46a648
    + WFrame: 0x55624d46a718
     + WGroupCW: 0x55624d60d628
      + WClientWin: 0x55624d619fc8
```
The repl.inspect_object if one of @wilhelmy repl helper functions runs up and down the object tree.
WRootWin.current now returns a WScreen without calling `rootwin_current_scr` :+1:

PR #283, opened 2020-05-28
Add more docs to cfg_notion.lua

raboof:


etc/cfg_notion.lua
```diff
@@ -55,14 +55,35 @@ ioncore.set{
...
+    -- Should notification tooltips be displayed for hidden workspaces with
+    -- activity?
+    --screen_notify=true,
+
+    -- Specifies where to add new regions on the mutually exclusive list of a
+    -- frame. One of ''last'', ''next'' (for after current), ''next-act''
+    -- (for after current and anything with activity right after it).
+    --frame_default_index="next",
+
+    -- Put transients in nested frames.
```
wilhelmy, line 70: I'm not sure what this means... Could you elaborate?

etc/cfg_notion.lua
```diff
@@ -55,14 +55,35 @@ ioncore.set{
...
+    --framed_transients=true,
+
     -- Float window placement method:
     -- one of 'udlr' (up down left right), 'lrud' (left right up down), 'pointer' or 'random'.
-    --float_placement_method="udlr"
+    --float_placement_method="udlr",
+
+    -- Auto-unsqueeze transients/menus/queries/etc.
```
wilhelmy, line 77: See above

etc/cfg_notion.lua
```diff
@@ -55,14 +55,35 @@ ioncore.set{
...
+    --unsqueeze=true,
 
     -- Float non-transient dialog type windows.
     -- Can be useful for applications that try to place their own dialog windows.
     --window_dialog_float=true,
 
+    -- Autoraise regions in groups on goto. \\
```
wilhelmy, line 84: I mean I appreciate brevity over unnecessary convolution but in these 3 cases I'm kind of lost, and I assume so is any user reading this.

raboof:
I just copied the documentation from https://github.com/raboof/notion/blob/master/ioncore/conf.c#L49 (which is also included in the docs).

I think it would be useful to expand on what these settings actually do, but then I think we should keep that in one place rather than duplicated.

I like having them in the default configuration. Perhaps we should change the docs to link to the configuration? Or even include it?

Issue #288, opened 2020-06-07
xinerama crashing bug on non-xinerama screens


kristopolous:
I was using a VNC session to do some diagnostics and I was getting a crash ... there's a null pointer issue, I fixed it, patch coming.

kristopolous:
see https://github.com/raboof/notion/pull/289

raboof:
Thanks!

PR #289, opened 2020-06-07
Xinerama crashing bug on non-xinerama screens

kristopolous:
There's a possibility of a null pointer returned by
XRRGetScreenResources in mod_xrandr_get_outputs_for_geom. It wasn't
getting checked. Now it is.


raboof:
Fixes #288

kristopolous:
I don't know tbh... I certainly don't have enough experience with this codebase to be able have and opinion on preemptive protections, it wasn't crashing out on me

raboof:
> I certainly don't have enough experience with this codebase to be able have an opinion on preemptive protections, it wasn't crashing out on me

Well, given that we now know `XRRGetScreenResources` can return null in some situations, I think we should make sure `mod_xrandr_get_all_outputs` doesn't crash in that situation ;). Filed #292

PR #294, opened 2020-06-14
Changing user facing verbiage of vertical horizontal to rows and columns

kristopolous:


etc/cfg_notion3_tiling.lua
```diff
@@ -39,9 +39,9 @@ defctxmenu("WTiling", "Tiling", {
     menuentry("Destroy frame",
               "WTiling.unsplit_at(_, _sub)"),
 
-    menuentry("Split vertically",
+    menuentry("Into rows",
```
raboof, line 42: Perhaps 'Split into rows' would be even clearer, but :+1:

PR #296, opened 2020-10-26
Remove some colorful language

raboof:
While I'll admit I got a chuckle out of such comments on occasion, and it's
fine to not be all 'professional' and have some fun in the comments, I can
do without the hateful undertone.


buhl (approved):
LGTM.

PR #302, opened 2020-12-04
Fix potential livelock in do_timer_set

### LINKED ISSUE #213: ion-statusd going to 100% cpu
Since upgrading to latest git, I've noticed ion-statusd going to 100% cpu. strace shows it receiving SIGALRM over and over again

reported by @trazillion on IRC

dnr:
The old code makes all timers interval timers, with the interval equal
to the initial delay, no matter how short the initial delay is (even
1µs). On some systems, this can cause so many signals to be delivered
that the code that resets the timer never gets to run.

The timer is theoretically one-shot; the interval part is only for
if we miss the signal due to a race condition. So we can use a fixed
small delay for the interval.

Fixes #213

### DIFF
--- a/libmainloop/signal.c
+++ b/libmainloop/signal.c
@@ -77,8 +77,10 @@ static void do_timer_set()
         return;
     }
 
-    val.it_interval.tv_usec=val.it_value.tv_usec;
-    val.it_interval.tv_sec=val.it_value.tv_sec;
+    // This really just needs to be a one-shot timer, but it is possible
+    // to miss it if it happens too soon. So add a small interval just
+    // so we _eventually_ wake up.
+    val.it_interval.tv_usec=100000;
 
     if((setitimer(ITIMER_REAL, &val, NULL))){
         had_tmr=TRUE;


libmainloop/signal.c
```diff
@@ -77,8 +77,10 @@ static void do_timer_set()
         return;
     }
 
-    val.it_interval.tv_usec=val.it_value.tv_usec;
-    val.it_interval.tv_sec=val.it_value.tv_sec;
+    // This really just needs to be a one-shot timer, but it is possible
+    // to miss it if it happens too soon. So add a small interval just
+    // so we _eventually_ wake up.
+    val.it_interval.tv_usec=100000;
```
raboof, line 83: Thanks for looking into this!

I understand how this helps avoid missing resetting the timer when the interval is too short. On the other hand, it will also shorten the timer to 100ms when it is actually scheduled for later. Perhaps we should only set tv_usec to 100000 when `val.it_value.tv_sec==0` and `val.it_value.tv_usec<100000`? And just to be sure explicitly set `val.it_interval.tv_sec=0` in that case?

dnr: I'm not an expert on this API, but I don't believe that's true: `it_value` is the time for the first period, and that hasn't changed. That is, if `it_value` is 15 seconds, and `it_interval` 100ms, then sigalrm will be fired once after 15 seconds, and then every 100ms after that, which I think is what we want.

`it_interval.tv_sec` is explicitly set to 0 at the top of the function, so we don't have to do it again.

raboof: ahh, right, I read that wrong. Looks good AFAICS!

raboof (approved):

PR #303, opened 2020-12-19
Initial implementation of _NET_FRAME_EXTENTS

### LINKED ISSUE #59: Firefox subwindows pop up on the wrong monitor
I have 3 monitors attached to my laptop: the internal screen, one to the left of it and one to the right.

When I keep firefox on the rightmost or center screen, some subwindows like context menus or other  popup menus (e.g. built-in dropdowns or input suggestions on previously filled forms but not ones which are implemented in javascript, I assume because they're not handled directly by GTK) appear at the correct position, but on the leftmost screen.

If I keep firefox on the leftmost screen, the popup works as expected.

It also doesn't happen with thunderbird or other GTK2 or GTK3 software, but affects all versions of firefox I've used on this machine up until 57.

Any ideas? Can anyone confirm this issue as being notion specific and not a firefox bug?

raboof:
Should fix #59

Co-Authored-By: Florian Larysch <fl@n621.de>

### DIFF
--- a/ioncore/netwm.c
+++ b/ioncore/netwm.c
@@ -37,7 +37,9 @@ static Atom atom_net_wm_moveresize=0;
 static Atom atom_net_wm_window_type=0;
 static Atom atom_net_wm_window_type_dialog=0;
 
-#define N_NETWM 9
+static Atom atom_net_frame_extents=0;
+
+#define N_NETWM 10
 
 static Atom atom_net_supported=0;
 
@@ -65,6 +67,8 @@ void netwm_init()
     atom_net_wm_moveresize=XInternAtom(ioncore_g.dpy, "_NET_WM_MOVERESIZE", False);
     atom_net_wm_window_type=XInternAtom(ioncore_g.dpy, "_NET_WM_WINDOW_TYPE", False);
     atom_net_wm_window_type_dialog=XInternAtom(ioncore_g.dpy, "_NET_WM_WINDOW_TYPE_DIALOG", False);
+
+    atom_net_frame_extents=XInternAtom(ioncore_g.dpy, "_NET_FRAME_EXTENTS", False);
 }
 
 
@@ -82,6 +86,7 @@ void netwm_init_rootwin(WRootWin *rw)
     atoms[6]=atom_net_active_window;
     atoms[7]=atom_net_wm_allowed_actions;
     atoms[8]=atom_net_wm_moveresize;
+    atoms[9]=atom_net_frame_extents;
 
     XChangeProperty(ioncore_g.dpy, WROOTWIN_ROOT(rw),
                     atom_net_supporting_wm_check, XA_WINDOW,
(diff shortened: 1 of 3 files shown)


ioncore/netwm.c
```diff
@@ -101,6 +106,10 @@ void netwm_init_rootwin(WRootWin *rw)
     xwindow_set_utf8_property(rw->dummy_win, atom_net_wm_name, p, 1);
 }
 
+void netwm_set_frame_extents(WClientWin* cwin, long* extents) {
+    XChangeProperty(ioncore_g.dpy, cwin->win, atom_net_frame_extents,
+                    XA_CARDINAL, 32, PropModeReplace, (uchar*)extents, 4);
+}
```
florolf, line 112: This is not a very safe API: The caller needs to know that they have to pass 4 values in `extents`. I'd rather make this explicit, which also allows hiding the gotcha that a protocol-level `CARD32` is represented a `long` in the xlib ABI, no matter the actual `sizeof(long)`, while the actual `CARD32` type varies in size:

```suggestion
void netwm_set_frame_extents(WClientWin *cwin, CARD32 left, CARD32 right, CARD32 top, CARD32 bottom)
{
    unsigned long data[4];

    data[0] = left;
    data[1] = right;
    data[2] = top;
    data[3] = bottom;

    XChangeProperty(ioncore_g.dpy, cwin->win, atom_net_frame_extents,
                    XA_CARDINAL, 32, PropModeReplace, (uchar*)data, 4);
}
```

florolf (approved):
Thank you for proposing this. I was reluctant to do this because it might cause other regressions while simultaneously fixing #59: Other toolkits also have (possibly differently-buggy-than-GTK) fallback paths that trigger when this property is absent (as it was previously), see for example [QT](https://sources.debian.org/src/qtbase-opensource-src/5.15.2+dfsg-2/src/plugins/platforms/xcb/qxcbwindow.cpp/?hl=623#L623).

Furthermore, who knows what implicit assumptions toolkits/programs make based on the actual value of the property. EWMH doesn't _really_ define what it is actually supposed to mean or what clients are allowed to deduce from it. It only really makes sense when making further assumptions about the window hierarchy one ends up with after reparenting and neither EWMH nor ICCCM make any strong claims about this as far as I can see.

All of this leads to the rather philosophical question what extents actually _are_ in a window manager that is less WIMPy than assumed by those specs, such as notion. The closest thing would probably be the offset added by `frame_managed_geom`, but I don't know how to properly propagate this into a client window in all situations. The proper place to set the property would probably be `clientwin_fitrep`, but this can be called with a multitude of different parameters and even multiple times for a single attach; I'm not sure how to handle this correctly.

So I guess the easy way out is to claim that since the specs are unclear, a fixed `(0, 0, 0, 0)` is as valid a setting for this property as any. I've been using this setting for a week now and have not experienced any glitches in the applications I usually use, so it might not even be a problem in practice.

Given that the presence of this property mainly seems to serve to switch between different assumptions about the shape of the window hierarchy in different toolkits, maybe adding a winprop kludge to enable/disable setting this property is probably more useful than agonizing about the actual value if problems ever crop up because of this.

raboof:
Tested it starts up and the first few screens of the activation dialogs work, but I don't have a license, so I didn't get past the activation dialog.

ioncore/netwm.c
raboof: I agree it would be better to pass in the 4 values individually (or in a dedicated data structure). I think I'd prefer to pass them in as regular `int`s and do the conversion to `long`/`CARD32` inside the function, though. Amended, WDYT?

florolf: That's fine, I guess that's more consistent with the rest of the codebase anyway.

PR #306, opened 2020-12-27
Use Lua 5.4 if it is found

### LINKED ISSUE #305: Add Lua 5.4 to the build system
It was released a while ago, would be nice to build against it on systems that have it.

wilhelmy:
This closes #305 once merged. Please test, but I don't see anything here https://www.lua.org/manual/5.4/manual.html#8 that we would probably trip over.

### DIFF
--- a/build/lua-detect.mk
+++ b/build/lua-detect.mk
@@ -31,7 +31,7 @@ lua_ver_extract = $(shell $1 -v 2>&1 | cut -f2 -d' ' | cut -f-2 -d.)
 PKG_CONFIG_ALL_PACKAGES:= $(shell pkg-config --list-all | cut -f1 -d' ')
 
 # these are in order of preference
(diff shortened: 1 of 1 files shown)


raboof:
nice! if I'm reading https://archlinux.org/packages/extra/x86_64/lua/ right they're already using 5.4

nixos doesn't have 5.4 yet (https://github.com/NixOS/nixpkgs/issues/99261), I'll look into that a bit so I can test ;)

raboof (approved):
seems to work just fine indeed :+1:

raboof:
@wilhelmy do you want to keep this WIP or is this ready to go?

wilhelmy:
No, I just figured we should test it somewhat extensively, but since master isn't due for release so soon, maybe now would be a good time to merge it.

wilhelmy:
I've been running this for a few days now and it seems stable. Anyone notice any breakage?

raboof:
Seems healthy here!

PR #327, opened 2021-04-10
Fix compilation on OpenBSD

c0dev0id:
Hello,
The variable name "stdout" is defined in stdio.h on OpenBSD already. In order to make mod_notionflux compile on this OS, it would be necessary to rename this variable.


raboof (approved):
:+1:

PR #328, opened 2021-04-10
Find lua versions without dot

c0dev0id:
Hello,
OpenBSD names lua executables lua51, lua52 ...
I'm not sure if other systems are doing this as well. But it won't hurt to add those to the auto detection logic.


raboof (approved):
haven't seen it anywhere else, but I agree it likely can't hurt

Issue #320, opened 2021-03-03
mod_sp: set default width/height of the scratchpad from cfg_sp.lua


wilhelmy:
Currently it's a compile-time constant that's set to 640/480 which is very small on modern setups.

My idea would be to add a `mod_sp.set{ default_width=800, default_height=600 }` function (and a corresponding `get` function) similar to `ioncore.{get,set}`, unless someone has a better idea what an API that sets module specific values should look like because module specific settings that are not attached to one of the object hierarchy objects are kind a bit of an outlier that currently does not exist in notion aside from core.

buhl:
I had a little fun with scratchpads lately. I basically reimplemented mod_sp in lua with some ekstra things on top  in #317.
I think we could move mod_sp to be intirerly implimented in lua without users being effected.

Actually I have been spending some time in the notion codebase generally and I think this could be done more places if we wanted to.

wilhelmy:
mod_sp as it is doesn't seem to do much, and seems buggy. I'm all in favour of replacing C by a more modern language where possible. Will look at the code some time.

PR #329, opened 2021-04-10
Allow to specify the scratchpad size from lua config

c0dev0id:
Hi,
this allows to set the scratchpad dimension from the mod_sp configuration file.
It should fix https://github.com/raboof/notion/issues/320.

This allows you to specify `mod_sp.set_size("1400", "800");` to set a scratchpad size.
If no size is specified the old hard coded size applies.

### DIFF
--- a/mod_sp/cfg_sp.lua
+++ b/mod_sp/cfg_sp.lua
@@ -2,6 +2,9 @@
 -- Ion mod_sp configuration file
 --
 
+-- Set default size for scratchpad
+mod_sp.set_size(1400, 800);
+
 defbindings("WScreen", {
     bdoc("Toggle scratchpad."),
     kpress(META.."space", "mod_sp.set_shown_on(_, 'toggle')"),
--- a/mod_sp/main.c
+++ b/mod_sp/main.c
@@ -51,6 +51,8 @@ char mod_sp_ion_api_version[]=NOTION_API_VERSION;
 
 /*{{{ Exports */
 
+static int sp_width = CF_SCRATCHPAD_DEFAULT_W;
+static int sp_height = CF_SCRATCHPAD_DEFAULT_H;
 
 static WRegion *create_frame_scratchpad(WWindow *parent, const WFitParams *fp,
                                         void *UNUSED(unused))
@@ -82,8 +84,8 @@ static WRegion *create_scratchws(WWindow *parent, const WFitParams *fp,
     par.szplcy=SIZEPOLICY_FREE_GLUE;
 
     par.geom_set=TRUE;
-    par.geom.w=MINOF(fp->g.w, CF_SCRATCHPAD_DEFAULT_W);
-    par.geom.h=MINOF(fp->g.h, CF_SCRATCHPAD_DEFAULT_H);
+    par.geom.w=MINOF(fp->g.w, sp_width);
+    par.geom.h=MINOF(fp->g.h, sp_height);
     par.geom.x=(fp->g.w-par.geom.w)/2;
     par.geom.y=(fp->g.h-par.geom.h)/2;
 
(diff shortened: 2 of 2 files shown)


mod_sp/cfg_sp.lua
```diff
@@ -2,6 +2,9 @@
 -- Ion mod_sp configuration file
 --
 
+-- Set default size for scratchpad
+mod_sp.set_size("1400", "800");
```
wilhelmy, line 6: hmm... wouldn't it be nicer to pass an integer here?

c0dev0id: Yes it would be nicer, and I tried. I'm not super familier with the notion code base yet. It looks like libextl-mkexports supports int, but when I try, it fails.

```
Scanning main.c for exports.
/usr/local/bin/lua53: ../libextl/libextl-mkexports:161: Error parsing type from "int* width"
stack traceback:
	[C]: in function 'error'
	../libextl/libextl-mkexports:22: in function 'errorf'
	../libextl/libextl-mkexports:161: in function 'parse_type'
	../libextl/libextl-mkexports:225: in upvalue 'do_do_export'
	../libextl/libextl-mkexports:301: in local 'minfn'
	../libextl/libextl-mkexports:365: in function 'do_parse'
	../libextl/libextl-mkexports:348: in function 'parse'
	../libextl/libextl-mkexports:821: in main chunk
	[C]: in ?
gmake[1]: *** [../build/rules.mk:55: exports.c] Error 1
gmake[1]: Leaving directory '/usr/ports/pobj/notion-4.0.2/notion-4.0.2/mod_sp'
gmake: *** [build/rules.mk:176: subdirs] Error 2

```

mod_sp/main.c
```diff
@@ -168,6 +170,19 @@ bool mod_sp_create_scratchpad(WScreen *scr)
...
 }
 
+/*EXTL_DOC
+ * Set default dimension for newly created scratchpads
+ * with parameters \var{width} and \var{height}.
+ */
+EXTL_EXPORT
+void mod_sp_set_size(int width, int height)
+{
+    if(width != NULL)
```
wilhelmy, line 180: technically with the change to int this should be `!= 0` but other than that I'd say LGTM. Thanks!

wilhelmy: Oh, and it is possible to pass negative values... `> 0`? :)

c0dev0id: Yes, I was a little trigger happy. No worries, things should be correct. Good work to call out the little bits!

mod_sp/main.c
```diff
@@ -177,9 +177,9 @@ bool mod_sp_create_scratchpad(WScreen *scr)
 EXTL_EXPORT
 void mod_sp_set_size(int width, int height)
 {
-    if(width != NULL)
+    if((width > 0) && (width < INT_MAX))
```
wilhelmy, line 180: Now I'm slowly reaching nitpick territory where it borders driving contributors away, but it's probably best to pick the nits early on ;')
Personal preference here: parentheses only when they make things clearer/serve a purpose. `width < INT_MAX` is redundant because an int can never exceed it. "Normal" notion coding style is to omit spaces where possible but I personally think that's crazy. However these are nits and no real objections towards merging from me.

c0dev0id: Fine for me. Every project has its own standards and it's good to stick to them (learn them).

wilhelmy (approved):
lgtm

raboof (approved):
Is it guaranteed the lua configuration is always evaluated before the scratchpads are created?

wilhelmy:
I'd hope so. After all workspace definitions are in the same lua files, no?

c0dev0id:
It's loaded on load and then saved with the layout.

So...
1. change size
2. restart notion
3. recreate scratchpad

Then it has the new default size. The same as before with the hard coded value. Or if there was no saved session the size take effect immediately.

PR #331, opened 2021-04-10
Remove bash dependency

c0dev0id:
Hello,
this PR removes the bash dependency to install notion. It also fixed an incompatibility, because bash is not always in `/bin/bash`. On BSDs it's usually under `/usr/local/bin`.

Besides that. There's nothing in these scripts that need the bash.

I didn't touch the docker scripts and also let `nextversion.sh` alone. I think those are maintainer tools.


raboof:
(superseded by #332)

PR #332, opened 2021-04-10
Remove bash dependency

c0dev0id:
Hello,
this change removes the dependency on bash to build notion.

I don't think where -o pipefail would be of use in this script. I changed it to -e so it would brake in case an unhandled error would occur. It runs fine here without bash installed.


raboof (approved):
Indeed nice. Let's fix any bash-isms that might be in there when we encounter them.

PR #333, opened 2021-04-11
Remove handling for external libtu/libextl

c0dev0id:
Hello,

libtu and libextl are bundled with notion. Is the handling of an external libtu/libextl still required? I don't think so, but YMMV.


raboof (approved):
LGTM

I'm not aware of someone relying on this. If we find out about that we can always re-introduce.

PR #338, opened 2022-02-25
statusbar_workspace.lua

chertyhansky:
Shows name of active workspace for two-screen configuration with mod_xinerama. Modification of legacy/statusbar_workspace.lua.

### DIFF
--- a/contrib/statusbar/statusbar_workspace.lua
+++ b/contrib/statusbar/statusbar_workspace.lua
@@ -1,3 +1,7 @@
+-- Authors: Rico Schiekel <fire@paranetic.de>, Canaan Hadley-Voth, Kevin Granade <kevin.granade@gmail.com>, Akaky Chertyhansky <akakychert@gmail.com>
+-- License: Unknown
+-- Last Changed: 2022-02-23
+
 -- statusbar_workspace.lua
 --
 -- Show current workspace name in the statusbar.
--- a/contrib/statusbar/statusbar_workspace.lua
+++ b/contrib/statusbar/statusbar_workspace.lua
@@ -1,0 +1,66 @@
+-- statusbar_workspace.lua
+--
+-- Show current workspace name in the statusbar.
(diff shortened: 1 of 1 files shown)


raboof (commented):
Sounds good! Under what license is this script contributed? Might be nice to make that explicit in the script?

chertyhansky:
In the original statusbar_workspace it was "License: Unknown", so I don't know may I publish it with GNU GPL 3.0, for example?

raboof:
> In the original statusbar_workspace it was "License: Unknown", so I don't know may I publish it with GNU GPL 3.0, for example?

That doesn't seem right - perhaps ping the authors to see if they're OK with releasing under the LGPL? Otherwise explicitly mentioning 'License: Unknown' might at least accurately reflect the state of affairs.

kevingranade:
For my part I'm fine with any free license (Non-exclusively GPL, LGPL, BSF, MIT).

raboof (approved):

wilhelmy:
I think `mod_xinerama.query_screens()` can be used to query the correct number of screens in case you want to make it work for an arbitrary number of screens.

```lua
lua> for k,v in pairs(mod_xinerama.query_screens()) do print(k,v) end
1	table: 0xa95c10
2	table: 0x943dd0
nil
lua> for k,v in pairs(mod_xinerama.query_screens()[1]) do print(k,v) end
w	1600.0
x	1920.0
y	0.0
h	900.0
nil
lua> for k,v in pairs(mod_xinerama.query_screens()[2]) do print(k,v) end
w	1920.0
x	0.0
y	0.0
h	1200.0
nil
```

wilhelmy:
@xfire we're currently debating the license this script should have. Any input from your side?

xfire:
@wilhelmy oh wow, that's from very long time ago. Just go with whatever suits you guys best as long as it's a free license. 👍

PR #339, opened 2022-04-07
Update README.md

kristopolous:
bumping the lua version in the readme to a more modern one for a painless debian install


raboof (approved):
of course the *most* painless Debian install is to use the [packages](https://packages.debian.org/search?keywords=notion&searchon=names&suite=all&section=all), but this is a good improvement for sure ;)

PR #341, opened 2022-06-11
Gentoo patches

SoapGentoo:
Hi @raboof 
I'd like to upstream the patches for `notion` we need in Gentoo.


raboof (approved):
changes look good, verified the build also still works for the Nix package, thanks!

libextl/system-autodetect.mk
```diff
@@ -65,24 +67,24 @@ LUAC=$(LUA_DIR)/bin/luac
 
 ifndef LUA_MANUAL
```
wilhelmy, line 69: Makes me wonder if this entire section isn't actually duplicated here from build/lua-detect.mk, libextl lives in-tree.

PR #342, opened 2022-06-13
Respect cflags

SoapGentoo:
@raboof the second part of our patches. Most of these should simplify the build system and allow us to override all variables.

I've put the rationale for removing `RANLIB` into the commit message: `ranlib` is also necessary when re-indexing static archives after stripping them.


raboof (approved):
seems to work for the nixos package as well, so :+1:

PR #343, opened 2022-06-13
Invoke `ar -s` for non-GNU/SysV toolchains

SoapGentoo:
@raboof some older BSD toolchains require `-s`, so this should fix it on those OSes too.

### DIFF
--- a/build/rules.mk
+++ b/build/rules.mk
@@ -3,7 +3,7 @@
 ##
 
 AR ?= ar
-ARFLAGS = cr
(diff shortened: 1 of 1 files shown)


SoapGentoo:
POSIX specifies that `ar cr` should be enough to generate the correct index, but I read that old BSD toolchains would require `ar -s` without which BSD `ar` wouldn't generate the index. On modern POSIX OSes, `-s` can be omitted, I just want this to work fine on older BSDs too.

PR #346, opened 2022-10-30
Parallel build fixes

trofi:
Two small fixes for missing generated headers:

      mod_mgmtmode: add main.o depend on exports.h
      mod_notionflux: add mod_notionflux.o depend on exports.h

Both fixes are derived from build failures found by `make --shuffle`
added upcoming GNU `make-4.4`.

Before the change `notion` build failed after 2-3 build attempts on missing
`exports.h` if any of two fixes above is not applied.
After the change `notion` survived 15 rebuilds.


raboof (approved):
Nice, thanks!

PR #353, opened 2023-02-15
Rescue the client window group instead of the client window on unsplitting

### LINKED ISSUE #316: Intermittent bug in notion 4: META-Return (full-screen) sometimes ignored
Hi. I recently moved to notion-4 fulltime, and it's 99% functional (thanks tons!) but periodically I see issues.

I haven't debugged this yet, but maybe just the description will speak to you. I have META-Return bound to toggle the full-screen state of a window:

```
defbindings("WGroupCW", {
    bdoc("Toggle client window group full-screen mode"),
    kpress(META.."Return", "WGroup.set_fullscreen(_, 'toggle')"),
})
```

This works, and I've been using it for at least 10 years. With notion-4 I sometimes see existing emacs windows stop being able to react to META-Return. All of a sudden notion stops interpreting that key combination, and passes the keystroke to the application (emacs in this case), and emacs then tells me it doesn't know what to do with it. Which is right: it wasn't supposed to see it; notion was supposed to handle it upstream. 

So far I've seen this happen only to emacs and only to META-Return. Other key combinations continue to work. Other emacs windows (same executable) work. Killing the broken window, and bringing up a new one is the only work around I've found so far. Obvious suspects?

C4tion:
Rescue the client window group instead of the client window on unsplitting.
Also small stuff I stumbled upon.

Someone who knows what they are doing should review this. It might fail spectacularly in some cases.
I have no clue why every client window is managed by a client window group.

Generating a graph of the objects and managing relation helped debugging the issue:
```
echo "digraph managing {$(notionflux << EOF | grep -v '^nil$'

function label(sub)
    return tostring(sub) .. (sub:name() and " (" .. sub:name() .. ")" or "");
end

function dump(sub)
    print("\"" .. label(sub:manager()) .. "\" -> \"" .. label(sub) .. "\";");
    return true;
end

notioncore.clientwin_i(dump)
notioncore.region_i(dump)

EOF
)}" | dot -Tpng -o /tmp/graph.png
```

fixes #316

### DIFF
--- a/ioncore/manage.c
+++ b/ioncore/manage.c
@@ -370,7 +370,16 @@ bool region_rescue_some_clientwins(WRegion *reg, WRescueInfo *info,
                     break;
                 }
             }
-            if(!pholder_attach(info->ph, 0, (WRegion*)cwin))
+
+            WRegion *movereg = (WRegion*)cwin;
+
+            /* If the client window is managed by a GroupCW, move the GroupCW instead. */
+            WGroupCW *cwg = REGION_MANAGER_CHK(movereg, WGroupCW);
+            if(cwg!=NULL){
+                movereg = (WRegion*)cwg;
+            }
+
+            if(!pholder_attach(info->ph, 0, movereg))
                 fails++;
         }
     }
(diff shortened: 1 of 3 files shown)


dkogan:
A quick test says that the cant-unfullscreen-window bug is gone. Thanks!
I'm not familiar-enough with the code to say anything more

raboof (approved):

PR #355, opened 2023-02-24
mod_notionflux: Add ifdef for fchmod on FreeBSD (closes #350)

### LINKED ISSUE #350: mod_notionflux: fchmod of socket always fails on FreeBSD
On FreeBSD, mod_notionflux fails on start (xmessage opens mentioning an invalid argument error) because of an invalid arg related to the listening socket.  `start_listening` tries to fchmod the newly created socket but on FreeBSD this always fails.  The manpage for fchmod says:

```
     [EINVAL]           The fd argument refers to a socket, not to a file.
```

The following (hacky) patch fixes the issue and mod_notionflux seems to start up fine.

```diff
diff --git a/mod_notionflux/mod_notionflux.c b/mod_notionflux/mod_notionflux.c
index 51949e07..e9176c7c 100644
--- a/mod_notionflux/mod_notionflux.c
+++ b/mod_notionflux/mod_notionflux.c
@@ -328,8 +328,10 @@ static bool start_listening()
     if(listenfd<0)
         goto errwarn;
 
+#ifndef __FreeBSD__
     if(fchmod(listenfd, S_IRUSR|S_IWUSR)<0)
         goto errwarn;
+#endif
 
     addr.sun_family=AF_UNIX;
 
```

wilhelmy:


wilhelmy:
I guess this code could be removed. It might also not work on other BSD derivatives, but it seems no one has tried. It's the only place in the codebase that uses the `__FreeBSD__` macro and fchmod on sockets that don't exist in the file system is probably not quite portable. Removing the code actually shouldn't add a race condition because although we're calling bind, since notion isn't multithreaded there is nothing actually reading from the socket before the function returns, so I think I'd rather remove it.

raboof:
> I guess this code could be removed. It might also not work on other BSD derivatives, but it seems no one has tried. It's the only place in the codebase that uses the `__FreeBSD__` macro and fchmod on sockets that don't exist in the file system is probably not quite portable. Removing the code actually shouldn't add a race condition because although we're calling bind, since notion isn't multithreaded there is nothing actually reading from the socket before the function returns, so I think I'd rather remove it.

If you're confident, then I'd be happy to merge a PR just removing it (perhaps adding a comment with the background, for the unlikely case we'd ever add multithreading here)?

PR #360, opened 2023-04-04
Add "acrobatic" kludge for the Ghidra SRE

### LINKED ISSUE #359: Strange behavior with Ghidra
I've been having a weird problem with Ghidra in notion for a while now.  The problem occurs when two Ghidra windows want to display in the same screen.  The way that Ghidra works is that if the main "CodeBrowser" window does not have focus, the other windows are not displayed.  I think this is what is happening:

1. I click on CodeBrowser to bring the CodeBrowser into focus
2. Ghidra displays the windows that were hidden
3. One of these windows takes the focus
4. Ghidra hides the window because the CodeBrowser is not in focus
5. The CodeBrowser now becomes in focus again
6. Go to step 2

Is there any way to avoid this?  I'm going to try playing with some kludges (Acrobatic)...

Recording 2023-04-04 at 10 03 39

edmcman:
Fixes #359

### DIFF
--- a/etc/cfg_kludges.lua
+++ b/etc/cfg_kludges.lua
@@ -45,6 +45,12 @@ defwinprop {
     acrobatic = true,
 }
 
+-- https://github.com/raboof/notion/issues/359
+defwinprop {
+    class = "ghidra-Ghidra",
+    acrobatic = true,
+}
+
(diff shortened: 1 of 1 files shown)


raboof:
Thanks!

PR #367, opened 2023-11-26
Create return placeholder for windows mapped as fullscreen

### LINKED ISSUE #366: Clients mapped in fullscreen state get additional WGroupCW when switched to windowed state
Steps to reproduce:
1. Create a window that is fullscreen from the start: `mpv bbb_sunflower_1080p_60fps_normal.mp4 -fs`
2. Toggle fullscreen with Notion keybinds.

Observed behavior:
The title of the windowed mpv is now `WGroupCW<...>`.

Expected behavior:
The title should be `bbb_sunflower_1080p_60fps_normal.mp4 - mpv`.

This might be related to #247, but the issue described there is not fixed by my patch.

C4tion:
Fixes #366.

`region_prepare_manage` did return a `WGroupedPHolder` while `region_make_return_pholder` returns a `WGroupPHolder`, which would also be created if we start in windowed state.

I am not sure why this works. It might break stuff that I do not use.

### DIFF
--- a/ioncore/manage.c
+++ b/ioncore/manage.c
@@ -150,8 +150,7 @@ static bool try_fullscreen(WClientWin *cwin, WScreen *dflt,
         fs_scr=clientwin_fullscreen_chkrq(cwin, param->geom.w, param->geom.h);
 
     if(fs_scr!=NULL){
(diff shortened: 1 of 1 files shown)


wilhelmy:
Sorry for letting this linger for so long. I'd like to test it sometime soon when I get around to it.

PR #374, opened 2025-01-14
Changes goto_previous to accept an integer number of which previous

kristopolous:
This addresses a long outstanding issue I've had ... I often want things like a basic alt+tab without a lot of cognitive load. goto_previous was fine if the desire is 2, but oftentimes it's 3 and the problem is then we're either managing window placement, doing directional cycle, which takes mental bookkeeping or doing the focus-list cycle which takes 1 keystroke to bring it up, then another to cycle through, another to select it. These are all very inconvenient.

So this is a patch to make goto_previous accept a number. If it's 1, then it works as before. If it's 2 then it cycles between 3 windows. If it's 3, then it cycles between the previous 4 and so on. 

It is a *breaking change* and I'll be proposing another PR that has a more complicated notion of both directionality and offset which will have a new function signature and not be breaking.

Feel free to accept this, the other, or neither.

### DIFF
--- a/ioncore/focus.c
+++ b/ioncore/focus.c
@@ -81,7 +81,7 @@ static void region_focuslist_deinit(WRegion *reg)
  * actually have received the focus when this function returns.
  */
 EXTL_EXPORT
-WRegion *ioncore_goto_previous()
+WRegion *ioncore_goto_previous(int count)
 {
     WRegion *next;
 
@@ -99,7 +99,7 @@ WRegion *ioncore_goto_previous()
         next!=NULL;
         next=next->active_next){
 
-        if(!REGION_IS_ACTIVE(next))
+        if(!REGION_IS_ACTIVE(next) && --count <= 0)
             break;
     }
 
--- a/ioncore/focus.h
+++ b/ioncore/focus.h
@@ -52,7 +52,7 @@ extern void region_focuslist_move_after(WRegion *reg, WRegion *after);
 /* Notify the focus handling that the region is deinit'ing */
 extern void region_focus_deinit(WRegion *reg);
 
-extern WRegion *ioncore_goto_previous();
+extern WRegion *ioncore_goto_previous(int count);
 
 /* Handlers to this hook should take WRegion* as parameter. */
 extern WHook *region_do_warp_alt;


mbenkmann:
I don't understand your gripe with the `focuslist`. My current binding is this:

`kpress("Mod1+Tab", "mod_menu.grabmenu(_, _sub, 'focuslist', {sizepolicy='center'})"),`

So switching to the previously focussed window is one keystroke only: Alt+Tab.  If I want to switch to the window focussed before the previously focussed one it's 2 keystrokes: Alt+Tab+Tab. 

I don't see how you could go lower in keystrokes than that?

And if you want to limit the choices or change the ordering, I suggest you write Lua code to maintain your own list of N choices and just call grabmenu() with your own list. I don't see a need for a change to the Notion base code at all to achieve what you want. It should be possible to do completely in Lua.

kristopolous:
> I don't understand your gripe with the `focuslist`. My current binding is this:


That was extremely dismissive, aggressive and hostile.

Is this how you intend to respond to people volunteering and trying to make things better?

mbenkmann:
Sorry for spamming this issue, but the accusation that being hostile has upset me, so I feel the need to explain myself.

> I don't understand your gripe with the focuslist.

When I wrote this, I meant it exactly this way: I don't understand. There is no hidding meaning, no implication that you are wrong or anything like that. Just the fact that I do not understand why you are dissatisfied with the focuslist.

> My current binding is this: ...

This and the following stuff I am writing to tell you how I use the focuslist. I am writing this so you know what I am talking about, because I think we might be talking about different things. I was not being condescending, not trying to tell you how you should do things. I was just telling you what *I* do, so that you know what I am talking about.

> I don't see how you could go lower in keystrokes than that?

Again, there is no hidding meaning here. Just the statement that I don't understand how one could make this more efficient. I am NOT saying that you are  wrong and it can't be made more efficient. But from your description of your PR I don't understand it. I am perfectly willing to accept that the problem in understanding is with me. But I want to understand how it can indeed be made more efficient.

> And if you want to limit the choices or change the ordering, I suggest you write Lua code to maintain your own list of N choices 

This should tell you clearly that even though I don't understand exactly why the existing functionality does not fit your requirements, I accept that it doesn't. I am suggesting to write an alternative to the existing focuslist because I accept that your issue is valid and cannot be met with the existing focuslist.

> I don't see a need for a change to the Notion base code at all to achieve what you want. It should be possible to do completely in Lua.

This was not intended as a criticisim of your code. It was just because pure Lua contributions have a higher chance of being accepted that I was suggesting going that route.

So again, please help  me understand why what I wrote upset you? It was certainly not my intention.

kristopolous:
ok so here's the use case:

Let's say you want to cycle between the 2 previous windows, 3 total. Yuur model is

`key + tab + tab` ... 

What this model allows you to do is set a cycle window length so Instead of one action cycling the current and the previous, it's of variable length ... let's put this in your terms

Press Mod1 + tab + tab + tab
Then repeat those 4 keystrokes a bunch of times

You've essentially established your working set being those 4 windows - you can make it 3, you can make it 6, make it what ever you want. It's a different way to group things without spacially grouping them.

So what I did was have a way to set this number based on my workflow then I cycled through the group without having to keep mental track of a stack in my head or even having to read a list,  

This is because sometimes you are working with things that need their own geometry - maybe window 1 is full screen, 2 is split, whatever ... 

Maybe there is a pure lua way of doing this and if I could have figured that out, I wouldn't have filed a ticket, I wouldn't have found it in the code and I wouldn't have done a PR.

mbenkmann:
Take a look at this:

[curated_focuslist.txt](https://github.com/user-attachments/files/21558323/curated_focuslist.txt)

If I understand correctly, this should give you the behaviour as if `goto_previous(2)` was being used. Even if it's not exactly what you want, it shows the approach I was talking about. It should be possible to tailor this further to meet your needs. There are also things with this approach that might be interesting to you that goto_previous(n) can't achieve. E.g. you could add a sort() to the entries. This would result in a consistent activation order based on the window title, regardless of the order in which the windows were previously activated. And the order would stay the same even if you switch to a window out of order.  I think this would go well with your theme of reducing mental load. Another idea would be to filter certain windows out of the list. Let's say you are working with application windows A,B and C, but sometimes you need to check your emails. You could make it so your mail program never shows up in the list. So even if you occasionally switch to your mail program, the modified Alt-Tab with the curated focuslist would always bring you back to your ABC rotation without replacing one of the 3 windows with the mail program.

PR #375, opened 2025-01-14
implements the ioncore-goto-hop alternative

kristopolous:
Here's the alternative as promised. It introduces a "goto-hop" which does not break anything but introduces a new function that can go both ways

I couldn't figure out a pure lua way to do this.


kristopolous:
After testing, this DOES NOT work as intended. What a pity. I'll think about this a little more.

